PR #29, opened 2016-07-28
feat: add `defaultInitialVersion` option

jviotti:
This option, which defaults to `0.0.1`, can be used to configure the
desired default initial version in case the
`getChangelogDocumentedVersions` hook doesn't find any version.

If the default version reference can't be found on git, then all commits
from the beginning of the project will be included in such version.

Change-Type: minor
Changelog-Entry: Add `defaultInitialVersion` option.
Signed-off-by: Juan Cruz Viotti <jviottidc@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,13 @@ possible by using template helpers, consider declaring the
 This option specifies the desired location of your `CHANGELOG` file, relative
 to the root of your project.
 
+### `defaultInitialVersion (String)`
+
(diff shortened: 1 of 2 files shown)


jviotti:
This is not ready to be merged yet. As it is, if the CHANGELOG file doesn't yet exist, the default initial version will be resolved, but it doesn't mean that such a tag exists.

PR #44, opened 2016-11-28
Alter versionist.conf.js to a standard resin.io format.

hedss:
This new template version is intended to be a forward-looking template for all future resin.io projects.

Connects to #43 

change-type: patch


versionist.conf.js
```diff
@@ -1,32 +1,62 @@
...
+  // Only include a commit when there is a footer of 'Change-Type'.
+  // Ensures commits which do not up versions are not included.
+  // Commit messages without a relevant footer will not be included in the CHANGELOG.
   includeCommitWhen: (commit) => {
-    return commit.footer['Changelog-Entry'];
+    commit.footer = lowerKeys(commit.footer);
+    return !!commit.footer['change-type'];
   },
 
+  // Determine the type from 'change-yype:' tag.
```
lurch: Typo: `'change-type:'`

lurch:
I wonder if this should be added as a `resinio-versionist.conf.js` file, given that versionist is intended (I assume) to have wider use beyond resin.io ?

hedss:
@lurch The discussion has been that `versionist` would take on the standard template we want to use going forward. This would mean that it, itself, would also adhere to the template. I don't think that precludes this template from becoming a default (and indeed an example) for any other project?

lurch:
Nope, not at all. Never hurts to have solid examples to learn from :+1: 

@jviotti Does this change impact on any of the examples in the documentation or the tests?

versionist.conf.js
```diff
@@ -1,32 +1,62 @@
...
+        commit.subject = commit.footer['changelog-entry'];
+      }
+    });
+
+    return data;
   },
 
   template: [
-    '## {{version}} - {{moment date "Y-MM-DD"}}',
+    '# v{{version}}',
```
lurch: Why remove the date? Always helpful to know how stale / fresh a piece of code is (even if it's on a separate line rather than the same line as the version string).

hedss: Fair enough, will update it!

hedss:
@lurch The test files are auto-generated, it does not impact them.

As an aside, the reason I hadn't yet assigned anyone to this is I've found a few other issues to look into before this and several components using it should really be reviewed. I really made the PR in a holding pattern, but there's currently no 'in-process' label for it. And that's actually an interesting point, @jviotti would you object to this repo including the 'standard' resin `flow/in-process`, `flow/in-review` etc. labels?

jviotti:
> would you object to this repo including the 'standard' resin flow/in-process, flow/in-review etc. labels?

Not at all, go for it!

jviotti:
> @jviotti Does this change impact on any of the examples in the documentation or the tests?

It shouldn't affect anything as far as I can tell. The tests create their own `versionist.conf.js`, and the [README example](https://github.com/resin-io/versionist#example) guides the user to checkout a specific branch, which is kept intact from these changes, although it should be updated at some point.

hedss:
I think we're ready for a review now, and I've added those people who've been involved with the versioning discussion. This is the proposed template for all components.

The one addition is that I've stuck with the `versionist` appending of entries after five lines. I'd like to use the same boilerplate at the top of all components that adopt this:
```
# Change Log

All notable changes to this project will be documented in this file.
This project adheres to [Semantic Versioning](http://semver.org/).

```
I think this very clearly, then, implies how versioning occurs.

What have I missed, what could be better?

lurch:
Is it worth adding a note to the boilerplate that the Changelog is automatically updated using versionist, and so shouldn't be manually edited?

versionist.conf.js
```diff
@@ -1,32 +1,62 @@
...
   },
 
-  updateVersion: 'npm',
-
+  // Only include a commit when there is a footer of 'change-type'.
+  // Ensures commits which do not up versions are not included.
+  // Commit messages without a relevant footer will not be included in the CHANGELOG.
   includeCommitWhen: (commit) => {
-    return commit.footer['Changelog-Entry'];
+    commit.footer = lowerKeys(commit.footer);
```
jviotti: What do you think about adding this lower case transformation inside Versionist itself? Maybe allow users to enable it with a boolean flag?

versionist.conf.js
```diff
@@ -1,32 +1,62 @@
...
+    return !!commit.footer['change-type'];
   },
 
+  // Determine the type from 'change-type:' tag.
+  // Should no explicit change type be made, then no changes are assumed.
   getIncrementLevelFromCommit: (commit) => {
-    return commit.footer['Change-Type'];
+    commit.footer = lowerKeys(commit.footer);
+    if (commit.footer['change-type']) {
```
jviotti: I believe this whole thing can be replaced with:

```js
return _.trim(_.get(commit.footer, 'change-type', '')) || undefined;
```

versionist.conf.js
CameronDiver: I agree, in the .conf.js file it was being called repeatedly on the same data.

lurch: Why a boolean? It could be tri-state: preserve-case/lower-case/upper-case.

hedss: I'm confused.
The lower-case transformation is to allow loose matching of the tag inside `versionist`, not to alter the tag in a commit or the entry that gets written to the CHANGELOG.
If you only want to match tags that preserve the case intact, then you'd simply ensure you didn't call 'lowerKeys' in the first place (ie. `lowerKeys: false`), as it'd simply skip tags that did not match the tag at all. Am I missing something here?

lurch: Ah, right. Apologies for misunderstanding (I was tired). IMHO `lowerKeys` should be `false` by default (if not specified).

hedss: Unfortunately, I don't agree. From previous experience, it can be easy to misremember the format of a tag; having to then go back and edit these, when the actual tag doesn't make it into a Changelog or `package.json` entry makes no sense to me.

I would rather the 'stricter' option be opt-in than opt-out.

hedss:
Am I going to take the comments from both here and https://github.com/resin-io/resin-builder/pull/51 to produce a unified change set for:

1. The default `versionist` configuration
2. Changes required to `versionist` from these conversations

Post-changes, I will create a new PR which, when merged, will make both this and the other referenced PR above obsolete, and closable.

versionist.conf.js
CameronDiver: I agree with Heds, if anything we should be going towards a more 'fuzzy' match for tags which is very accepting to reduce headaches when somebody doesn't use a capital in the correct place etc.

CameronDiver:
I would suggest that versionist passes the git commit hash to the various *.conf.js functions, so that they can then call directly into git to find out any information necessary. For me this was getting the author, but with the commit hash and the git shell command there's not a lot that can't be grokked.

versionist.conf.js
```diff
@@ -1,32 +1,63 @@
-module.exports = {
+var _ = require('lodash');
+var execSync = require('child_process').execSync;
 
-  subjectParser: 'angular',
+var getAuthor = (commitHash) => {
+  var result = execSync(`git show ${commitHash}`, { encoding: 'utf8' });
```
CameronDiver: Using `git show --quiet --format="%an" [hash]` will return only the author's name, so no regex required.

CameronDiver: Using `git show --quiet --format="%an" [hash]` will return only the author's name, so no regex required.

versionist.conf.js
```diff
@@ -1,32 +1,59 @@
-module.exports = {
+var _ = require('lodash');
+var execSync = require('child_process').execSync;
 
-  subjectParser: 'angular',
+var getAuthor = (commitHash) => {
+  return execSync(`git show --quiet --format="%an" ${commitHash}`, { encoding: 'utf8' }).replace('\n', '');
+}
```
lurch: Much like you did with `lowerFooterTags`, would it be worth including this as a standard option within versionist itself, which you could turn on or off with a flag in `versionist.conf.js`?

versionist.conf.js
```diff
@@ -1,32 +1,59 @@
...
+  // Always add the entry to the top of the Changelog, below the header.
   addEntryToChangelog: {
     preset: 'prepend',
-    fromLine: 5
+    fromLine: 6
   },
 
-  updateVersion: 'npm',
-
+  // Only include a commit when there is a footer of 'change-type'.
```
lurch: "... a footer of 'change-type' or a footer of 'changelog-entry'." ?
(unless I've misunderstood again)

And should it be "a footer **tag** of 'change-type'" ?

versionist.conf.js
```diff
@@ -1,32 +1,59 @@
...
+      commit.subject = commit.footer['changelog-entry'] || commit.subject;
+      commit.author = getAuthor(commit.hash);
+    });
+
+    return data;
   },
 
   template: [
-    '## {{version}} - {{moment date "Y-MM-DD"}}',
+    '# v{{version}} - {{moment date "Y-MM-DD"}}',
```
lurch: Just out of curiosity, what's the reason for the change in the number of `#` characters? (i.e. the header-level)

versionist.conf.js
hedss: Personally, I don't think so. Given a commit hash, users of `versionist` can devise however they want to use it.
Others may well want just the email address, or both name and email address, or the time of the commit, or... where would we stop?
Making case flexible is more binary.

versionist.conf.js
```diff
@@ -1,32 +1,59 @@
...
     '',
     '{{#each commits}}',
-    '{{#with footer}}',
-    '- {{capitalize changelog-entry}}',
-    '{{/with}}',
+    '{{#if this.author}}',
+    '* {{capitalize this.subject}} [{{this.author}}]',
+    '{{else}}',
+    '* {{capitalize this.subject}}',
+    '{{/if}}',
```
hedss: It certainly can. However I like obvious, atomic steps.

versionist.conf.js
hedss: Oh, you've pre-empted the response I've just given. Yes, see above.

versionist.conf.js
hedss: Good catch, yup.

versionist.conf.js
```diff
@@ -1,32 +1,59 @@
...
   },
 
-  updateVersion: 'npm',
-
+  // Only include a commit when there is a footer tag of 'change-type'
+  // or 'changelog-entry'.
+  // Ensures commits which do not up versions are not included.
   includeCommitWhen: (commit) => {
-    return commit.footer['changelog-entry'];
+    return !!commit.footer['change-type'] || !!commit.footer['changelog-entry'];
```
lekkas: Won't this result in *either* `change-type` *or* `changelog-entry` tagged commits to end up into the changelog? I thought the plan was to use `change-type` commits *only* in changelog entries, with their subject being the default changelog entry unless a `changelog-entry` type tag is added as well, which will override the commit subject.

hedss: Yep, I've mixed my notes up. :-/ For some reason the ones that ended up on the TODO list included single changelog entry commits, but having gone back, the call notes do indeed specify a commit must have both (if there's a `changelog-entry` at all).
Sorry, there's been so much changed over the last two weeks it's been tricky keeping everything aligned.

hedss:
I've managed to completely stuff up the rebase. Because of this, and because I want to get this done ASAP, I've created a new PR https://github.com/resin-io/versionist/pull/52 which is the reviewed changes. I'm going to close this one.

Issue #46, opened 2016-11-29
`--config` parameter throws error


hedss:
Using a `versionist.conf.js` configuration file in the same directory as running `versionist` returns expected results. However explicit use of the same configuration file by using the `--config` option throws an error:
```
argv.config.getChangelogDocumentedVersions(argv.config.changelogFile, callback);
                    ^    
    TypeError: argv.config.getChangelogDocumentedVersions is not a function
        at async.waterfall (/Work/tmp/versionist/bin/cli.js:102:17)
        at nextTask (/Work/tmp/versionist/node_modules/async/dist/async.js:5070:14)
        at Object.waterfall (/Work/tmp/versionist/node_modules/async/dist/async.js:5073:5)
        at Object.<anonymous> (/Work/tmp/versionist/bin/cli.js:99:7)
        at Module._compile (module.js:556:32)
        at Object.Module._extensions..js (module.js:565:10)
        at Module.load (module.js:473:32)
        at tryModuleLoad (module.js:432:12)
        at Function.Module._load (module.js:424:3)
        at Module.runMain (module.js:590:10)
```

PR #47, opened 2016-11-30
fix: ensure a specified config file is parsed correctly

hedss:
Due to the way `yargs` operates, a named option cannot be used
as an object to pass back a new set of options (in this case
the loaded configuration), as the key on `argv` essentially already
exists.

To mitigate this, the returned object is renamed `configuration` and
this is used in place of `config` for all further operations.

Connects to #46.

Change-Type: minor
Changelog-Entry: Ensure a specified config file is parsed correctly.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -56,7 +56,7 @@ const argv = yargs
   .config('config', 'configuration file', (file) => {
     try {
       return {
-        config: configuration.parse(configuration.load(file))
+        configuration: configuration.parse(configuration.load(file))
       };
     } catch (error) {
       showErrorAndQuit(error);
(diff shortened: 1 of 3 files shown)


jviotti:
Very, very nice! LGTM :+1:

lurch:
Just out of curiosity: why was this "connects to #46" rather than "fixes #46" ?

hedss:
Habit.
I like to close issues manually so I can satisfy myself that everything required has been addressed; gives me closure, quite literally, of an issue. :)

lurch:
Fair enough :-)

lurch:
Now that it's been merged, should the "in-review" label be removed?

hedss:
I guess, though the PR is closed.

Issue #48, opened 2016-12-06
There is no `lowerCaseFooterTags` option


hedss:
This would be extremely useful to standardise `versionist.conf.js` files, so that parsed footer tags can always be checked in a lower case format, ie:
```
  includeCommitWhen: (commit) => {
    return commit.footer['changelog-entry'];
  },
```
Which allows developers not to have to remember the exact case of footer tags (I frequently misremember `Changelog-Entry` as `ChangeLog-Entry`).

PR #49, opened 2016-12-06
Add a `lowerCaseFooterTags` option (default is `true`) which lowercases all footer tags keys

hedss:
This allows developers to mix case for tags and simplify checks for tags in the `versionist.conf.js` configuration file.
Alters the current test-suite so that test configs use lowercased tags.

Connects to #48.

Change-Type: minor


hedss:
I added this as a `minor`, though I'm now wondering if actually it's a `major`, as any pre-existing `versionist.conf.js` that looks for tags in mixed-case will either need to set `lowerCaseFooterTags: false` or alter the tag strings to match.

Thoughts?

CameronDiver:
Given that we're talking about Kbs of data max on a once-per-PR run cycle, I'd probably recommend providing both the unaltered tags and the lower case tags, preserving all existing workflows but adding a new consistent way of accessing the tags. E.g. when 
`lowerCaseFooterTags = true`
then
`footer['change-type'] = footer['Change-Type'] = ...`

hedss:
I'd actually considered that, but I wasn't sure about how confusing that might be, especially for cases where tags are _.forEach()d or the like.
Maybe it's not a bad idea after all, given you've echoed it.

CameronDiver:
I guess it depends on what the default for the `lowerCaseFooterTags` is; if true then it's probably not the best idea for what you've just mentioned, if it's false by default then the assumption would be that in the documentation for the `lowerCaserFooterTags` switch would show that it returns both of the tags, one unchanged, one lower case.

hedss:
It's `true` by default, which I believe it should be. And actually, if you're looping through tags, then you're probably looking for a particular case anyway. I think it's actually safe to do.

hedss:
Post comments, changes have been made to return both versions of a footer tag, should the tag not already be fully lowercase. I believe this solves backwards compatibility issues and gives a more open approach to searching for tags in the footers.

tests/cli/cases/no-lower-footer-tags.js
```diff
@@ -0,0 +1,81 @@
...
+shelljs.rm('-rf', TEST_DIRECTORY);
+shelljs.mkdir('-p', TEST_DIRECTORY);
+shelljs.cd(TEST_DIRECTORY);
+
+utils.createVersionistConfiguration([
+  '\'use strict\';',
+  'module.exports = {',
+  '  subjectParser: \'angular\',',
+  '  editVersion: false,',
+  '  lowerCaseFooterTags: false,',
```
jviotti: Maybe worth testing that when `lowerCaseFooterTags` is false, accessing the tags in lowercase yields nothing?

lib/tags.js
```diff
@@ -87,19 +88,25 @@ exports.parseTagLine = (line) => {
...
     })
     .map((tagLine) => {
       const tag = exports.parseTagLine(tagLine);
+      const finalTags = [];
 
-      return [
-        tag.key,
-        tag.value
-      ];
+      finalTags.push([ tag.key, tag.value ]);
```
jviotti: I haven't actually test this, but I believe this whole thing can be simplifies as follows:

```js
.map((tagLine) => {
  const tag = exports.parseTagLine(tagLine);

  if (lowerCaseFooterTags) {
    tag.key = tag.key.toLowerCase();
  }

  return [
    tag.key,
    tag.value
  ];
})
```

This also has the advantage that `.flatten()` is not needed anymore.

If you prefer a more concise functionally-pure style:

```js
.map((tagLine) => {
  const tag = exports.parseTagLine(tagLine);

  return [
    lowerCaseFooterTags ? tag.key.toLowerCase() : tag.key,
    tag.value
  ];
})
```

lib/tags.js
```diff
@@ -73,6 +73,7 @@ exports.parseTagLine = (line) => {
  * @public
  *
  * @param {String[]} footerTagLines - footer tag lines
+ * @param {Boolean} lowerCaseFooterTags - lowercase footer tag keys
```
jviotti: Can we accept an options object that contains this key instead, since the options passed to this function is likely to grow more in the future?

jviotti:
Can we add unit tests for `parseFooterTags()` that cover this functionality?

lib/tags.js
jviotti: Oh, I see now that your intention is to return both of them, so my comments above don't apply very much. I'll update them in a minute...

jviotti: So in that case a `map` no longer makes sense, since the intention is to return either one item or two per tag line. What about something like this?

```js
exports.parseFooterTagLines = (footerTagLines, lowerCaseFooterTags) => {
  return _.chain(footerTagLines)
    .reject((line) => {
      return _.isEmpty(line.trim());
    })
    .reduce((tags, tagLine) => {
      const tag = exports.parseTagLine(tagLine);

      tags.push([
        tag.key,
        tag.value
      ]);

      if (lowerCaseFooterTags) {
        tags.push([
          tag.key.toLowerCase()
          tag.value
        ]);
      }

      return tags;
    }, [])
    .fromPairs()
    .value();
};
```

There is no need to check for duplicates since `_.fromPairs()` will take care of that automatically:

```js
> const lodash = require('lodash');
> lodash.fromPairs([ [ 'a', 1 ], [ 'a', 1 ] ]);
{ a: 1 }
```

hedss: Sounds good!

lib/tags.js
hedss: Sure.

lib/tags.js
hedss: Done.

lib/tags.js
hedss: Done.

lib/tags.js
```diff
@@ -73,6 +73,8 @@ exports.parseTagLine = (line) => {
  * @public
  *
  * @param {String[]} footerTagLines - footer tag lines
+ * @param {Object} [options={}] - options
+ * @param {Boolean} options.lowerCaseFooterTags - lowercase footer tag keys
```
jviotti: Do you think we should provide an optional value here? As it is now, this option is actually "optional", since if its not passed, its considered falsy. If that's OK, we can edit the JSDoc annotation to:

```js
@param {Boolean} [options.lowerCaseFooterTags=false] - lowercase footer tag keys
```

Otherwise, we should set:

```js
@param {Boolean} [options.lowerCaseFooterTags=true] - lowercase footer tag keys
```

and set a default below:

```js
_.defaults(options, {
    lowerCaseFooterTags: true
});
```

jviotti:
Looking great! I just left a nitty-gritty JSDoc comment, but will approve already.

hedss:
Cheers @jviotti will shore up this then merge. :)

README.md
```diff
@@ -251,8 +288,8 @@ Given `parseFooterTags: true`:
   "subject": "My first commit",
   "body": "Lorem ipsum dolor sit amet",
   "footer": {
-    "Closes": "#1",
-    "Foo": "bar"
+    "closes": "#1",
+    "foo": "bar"
```
lurch: As per the previous example, shouldn't this now include `Closes`, `Foo`, `closes` & `foo` ?

Given that `lowerCaseFooterTags` presumably has no effect when `parseFooterTags` is false, would it have made sense to document `lowerCaseFooterTags` *after* `parseFooterTags` ?

lib/tags.js
```diff
@@ -73,6 +73,8 @@ exports.parseTagLine = (line) => {
  * @public
  *
  * @param {String[]} footerTagLines - footer tag lines
+ * @param {Object} [options={}] - options
+ * @param {Boolean} [options.lowerCaseFooterTags=false] - lowercase footer tag keys
```
lurch: Huh, how come the default is `false` here, when earlier it said the default was `true`?

README.md
hedss: Yes, that wasn't updated on the change to multiple tags, the README.md should read:
```
    "Closes": "#1",
    "closes": "#1",
    "Floses": "#1",
    "floses": "#1",
```

lib/tags.js
lurch: Ahh, thanks for the clarification :-)

Issue #50, opened 2016-12-06
Versionist should provide commit hash to config file


CameronDiver:
To gain more information as necessary, the config file should be provided with a commit's hash with other information

PR #51, opened 2016-12-06
Add the passing of commit hash to the config file.

CameronDiver:
The hash of the commit is retrieved by git, and parsed from the yaml, then passed through as normal to the configuration file.

Change-Type: minor
Connects-To: #50

### DIFF
--- a/lib/git-log.js
+++ b/lib/git-log.js
@@ -38,7 +38,8 @@ const tags = require('./tags');
  * See https://git-scm.com/docs/pretty-formats
  */
 exports.GIT_LOG_YAML_PRETTY_FORMAT = [
-  '- subject: >-',
+  '- hash: %H',
+  '  subject: >-',
   '    %s',
   '  body: |-',
 
@@ -204,11 +205,15 @@ exports.parseGitLogYAMLOutput = (output, options = {}) => {
       .join('\n')
       .value();
 
+    // The commit object to be returned
+    const commitObj = {
+      subject: options.subjectParser(commit.subject),
+      hash: commit.hash
+    };
+
     if (!options.parseFooterTags) {
-      return {
-        subject: options.subjectParser(commit.subject),
-        body: options.bodyParser(commit.body)
-      };
+      commitObj.body = options.bodyParser(commit.body);
+      return commitObj;
     }
 
     // We iterate on the commit body lines in reverse, and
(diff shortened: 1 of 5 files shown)


CameronDiver:
Tests also need to be changed as a lot of them don't expect to be also retrieving a hash.

lib/git-log.js
```diff
@@ -38,7 +38,9 @@ const tags = require('./tags');
  * See https://git-scm.com/docs/pretty-formats
  */
 exports.GIT_LOG_YAML_PRETTY_FORMAT = [
-  '- subject: >-',
+  '- hash: >-',
```
jviotti: The hash is a single string, so we shouldn't be using YAML multi-string support here (the `>-`). Something like `- hash: %H` should suffice.

jviotti:
> Tests also need to be changed as a lot of them don't expect to be also retrieving a hash.

Yep. Lets also add some tests for this functionality as well.

lib/git-log.js
CameronDiver: Ah that makes sense :) I've not really used yaml much.

tests/git-log.spec.js
```diff
@@ -183,6 +183,7 @@ describe('GitLog', function() {
             'Currently we had AppImage scripts and other resources in various',
             'different places in the code base.'
           ].join('\n'),
+          hash: undefined,
```
jviotti: Even better: why not updating the `utils.formatCommit` function to also accept a hash, and check that is propagated correctly?

lib/git-log.js
```diff
@@ -38,7 +38,9 @@ const tags = require('./tags');
  * See https://git-scm.com/docs/pretty-formats
  */
 exports.GIT_LOG_YAML_PRETTY_FORMAT = [
-  '- subject: >-',
+  '- hash:',
+  '    %H',
```
jviotti: Is there any reason why `%H` is in another line? Now that we're not using the multi line modified, this can be simply `- hash: %H`.

jviotti:
I left a minor comment in the YAML string, but will approve already :+1:

tests/utils.js
```diff
@@ -26,19 +26,22 @@ const indentString = require('indent-string');
...
  *
  * @param {Object} options - options
+ * @param {String} options.hash - commit hash
  * @param {String} options.subject - commit subject
  * @param {String} options.body - commit body
  * @returns {String} formatted commit
  *
  * @example
  * const formattedCommit = utils.formatCommit({
+ *   hash: b8ca72ee58bddc9d174b71615aaa0159849b2aca
```
lurch: Example hash here is in lowercase, but all the examples have their hashes in uppercase?

lurch:
Maybe a silly question, but is it worth adding a unit-test that the hash retrieved from git is in the correct format? (only hex-chars, specific length, all lowercase, etc.)

This PR doesn't include any modifications to the documentation - should it?

Only a nitpick, but I guess "then passed through as normal to the configuration file." in the commit-message actually means "then made available as an attribute on the `commit` objects in the configuration file." ?

CameronDiver:
I've added documentation and a single object in the parse function which is updated and returned as necessary. Also added into the readme information about the new commit object which is provided to the conf file. Testing the output of the git hash doesn't really make sense to me as it doesn't matter what git gives us back as long as it can address the commits. I agree with you on the commit message, I'll do a rebase before merge and make it more appropriate.

lurch:
On a side note, I guess it might be useful if the [documentation](https://github.com/resin-io/versionist/blob/master/README.md#template-string) explicitly stated what attributes are available in the `commit` objects given to the template, rather than simply 'inferring' it from the [parseFooterTags](https://github.com/resin-io/versionist/blob/master/README.md#parsefootertags-boolean) docs.

PR #52, opened 2016-12-09
Add a default resin.io `versionist` configuration file, for use in future projects

hedss:
Connects to #43

change-type: minor


versionist.conf.js
```diff
@@ -1,32 +1,58 @@
...
   includeCommitWhen: (commit) => {
-    return commit.footer['changelog-entry'];
+    return !!commit.footer['change-type'];
   },
 
+  // Determine the type from 'change-type:' tag.
+  // Should no explicit change type be made, then no changes are assumed.
   getIncrementLevelFromCommit: (commit) => {
-    return commit.footer['change-type'];
+    return _.trim(_.get(commit.footer, 'change-type', '')) || undefined;
```
lurch: Just out of curiosity: Does the `includeCommitWhen` function mean that this function will never return `undefined` ?

hedss: No. I spent some time looking at this originally when I first started using `versionist`. `getIncrementLevelFromCommit()` is always called, regardless. The `includeCommitWhen()` method determines whether something should be included for entries, not what the `semver` level is.

There was discussion on this here (it is a very long thread):
https://www.flowdock.com/app/rulemotion/r-process/threads/-iNMo_7zoYUgr4xcQkCH4wvPJ6a

lurch: Yeah, thinking about I guess that makes sense, and it's only "coincidence" that we're using the same filter for both of those functions. I guess theoretically another project might choose to only show changelog entries for commits by a particular author, but they'd still want the version to be bumped as appropriate for commits by _all_ authors.

Issue #53, opened 2016-12-09
Default versionist.conf.js incorrectly assumes local CWD


hedss:
When run in a non-versionist component, the `versionist.conf.js` attempts to pull in `lodash` from the current CWD. This is not where `versionist` lives, however, so the pull fails.

PR #54, opened 2016-12-09
Ensure that referencing lodash occurs via the versionist base directory and not the CWD

hedss:
…ry and not the CWD.

Connects to #53.

change-type: patch

### DIFF
--- a/versionist.conf.js
+++ b/versionist.conf.js
@@ -1,4 +1,4 @@
-var _ = require('lodash');
+var _ = require(`${__dirname}/node_modules/lodash`);
 var execSync = require('child_process').execSync;
 
 var getAuthor = (commitHash) => {


versionist.conf.js
```diff
@@ -1,4 +1,4 @@
-var _ = require('lodash');
+var _ = require(`${__dirname}/node_modules/lodash`);
```
lurch: Perhaps `${__dirname}/node_modules` could be stored in a 'global variable' that versionist itself supplies to `versionist.conf.js` ?

hedss: Yes, possibly, but given I've currently a broken published version, I'd really like to fix it.

hedss: Created #55 to ensure this doesn't vanish.

jviotti: I suggest to make Versionist accept a function being returned from `versionist.conf.js`, and if it detects that, pass lodash directly to it, just for convenience. For example:

```js
// versionist.conf.js

module.exports = (utils) => {
    // utils.lodash.isEmpty()
    // utils.lodash.each()
    // etc ...
    return {
        // The usual configuration
    };
};
```

Issue #58, opened 2016-12-14
Default to internal versionist.conf.js if none is passed


lekkas:
It might make sense to default to the internal `versionist.conf.js` configuration if none is passed, with the ability to override it with `-c`. This is similar to how `resin-lint` works (https://github.com/resin-io/resin-lint/blob/master/lib/resin-lint.coffee#L9)

hedss:
I think there's an ordering caveat here:
* If the `-c` option exists, use the referenced configuration file
* If no `-c` option, but there is `versionist.conf.js` in the CWD, use that (which is the current default configuration)
* If no `-c` option, and there is no `versionist.conf.js` in the CWD, fallback to the internal configuration

lekkas:
@hedss you're right, this precedence sounds good to me

lekkas:
cheers @craig-mulligan , I'm closing this

PR #61, opened 2016-12-19
Default to internal versionist.conf.js if none is passed (#58)

craig-mulligan:
Connects to #58.

change-type: minor

### DIFF
--- a/lib/cli/configuration.js
+++ b/lib/cli/configuration.js
@@ -22,6 +22,7 @@
 
 const _ = require('lodash');
 const presets = require('../presets');
+const fs = require('fs');
 
 /**
  * @summary Configuration default values
@@ -324,4 +325,22 @@ exports.load = (file) => {
 
     throw error;
   }
+};
+
+/**
+ * @summary Returns path of first existing file
+ * @function
+ * @public
+ *
+ * @param {array} paths - file paths for files that may or may not exist
+ * @returns {string} file path
+ *
+ * @example
+ * const configPath = configuration.firstExistingFile([ './foo.js', './bar.js' ]);
+ * console.log(configPath);
+ */
+exports.firstExistingFile = (paths) => {
+  return paths.find((p) => {
+    return fs.existsSync(p);
+  });
 };
--- a/tests/cli-configuration.spec.js
+++ b/tests/cli-configuration.spec.js
@@ -19,6 +19,10 @@
 const _ = require('lodash');
 const m = require('mochainon');
 const configuration = require('../lib/cli/configuration');
+const packageJSON = require('../package.json');
+const path = require('path');
+
+const internalConfPath = path.join(__dirname, `../${packageJSON.name}.conf.js`);
 
 describe('CLI Configuration', function() {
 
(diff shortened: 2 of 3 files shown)


lurch:
I guess a matching unit-test would be nice? ;)

pcarranzav:
Might it be better to do the catch -> load internal outside of the `exports.load` function? i.e. from where the function is called in cli.js - imo that would keep `exports.load` cleaner and falling back to the default can be done at a higher level.

(Not such a strong opinion though)

hedss:
Concur with Pablo, best to throw the error up and catch it in `cli.js`... main reason is, if on the offchance Versionist's own `versionist.conf.js` doesn't exist (and I realise that's unlikely), we're into rabbit-hole-land™. :-)

craig-mulligan:
Like this @hedss @pcarranzav?

craig-mulligan:
Actually, I think I'd prefer to move `configLoader` to the `configuration` module and then make the current `config.load` private. (renaming them accordingly) Any aversion to that?

hedss:
No objection here!

craig-mulligan:
Could you take another look @hedss?

tests/cli-configuration.spec.js
```diff
@@ -265,4 +266,14 @@ describe('CLI Configuration', function() {
 
   });
 
+  describe('.load()', function() {
+
+    it(`should load defualt internal ${packageJSON.name}.conf.js if file is not found`, function() {
```
lurch: `default` typo.

Also, does this change mean that if the config-file that the user specifies doesn't exist, versionist will silently load its internal config file, rather than error that the user specified a non-existent filename? IMHO that'd be undesired.

hedss: I think that's a fair point. If we're specifying a config file and it doesn't exist, we should error. If no config file is specified, and there's no `versionist.conf.js` in the CWD, then we should fall back to `versionist`s internal config.

jviotti: Yeah, I agree. Lets bail out if the user specifies a configuration file but it doesn't exist.

craig-mulligan:
I've amended so it to set's the yargs --config default after testing which files exist on the fs, appears to be a much cleaner approach. What you do guys think?

lib/cli/configuration.js
```diff
@@ -325,3 +326,21 @@ exports.load = (file) => {
...
+ * @public
+ *
+ * @param {array} fallbacks - file paths for files that may or may not exist
+ * @returns {string} file path
+ *
+ * @example
+ * const configPath = configuration.getDefault([ './foo.js', './bar.js' ]);
+ * console.log(configPath);
+ */
+exports.getDefaultPath = (fallbacks) => {
```
lurch: Maybe a bit nit-picky, but I'm not sure `getDefaultPath` is such a great name? a) it sounds a bit too generic and b) to me a "path" is usually a directory rather than a file.
This function is basically `getFirstExistingFilename` ;-)  (which I agree is a bit of a mouthful - open to other suggestions)

lib/cli/configuration.js
```diff
@@ -325,3 +326,21 @@ exports.load = (file) => {
     throw error;
   }
 };
+
+/**
+ * @summary Returns defualt configuration file path
```
lurch: Same misspelling of `default` as last time ;)

lib/cli/configuration.js
```diff
@@ -325,3 +326,21 @@ exports.load = (file) => {
...
+ * @function
+ * @public
+ *
+ * @param {array} fallbacks - file paths for files that may or may not exist
+ * @returns {string} file path
+ *
+ * @example
+ * const configPath = configuration.getDefault([ './foo.js', './bar.js' ]);
```
lurch: example usage doesn't actually match the name of the function.

tests/cli-configuration.spec.js
```diff
@@ -15,10 +15,12 @@
  */
 
 'use strict';
-
 const _ = require('lodash');
 const m = require('mochainon');
 const configuration = require('../lib/cli/configuration');
+const packageJSON = require('../package.json');
+const path = require('path');
+const internalConfPath = path.join(__dirname, `../${packageJSON.name}.conf.js`);
```
lurch: Is it worth inserting a blank line above to visually separate this from the `require` lines?

lib/cli/configuration.js
craig-mulligan: To me *how* we get the default path might change, this way the names always approps. But no strong feels.

craig-mulligan: Yea but calling it'd you'd always do smthing like `config.getDefualtPath()` so it'd be implied. Anyway after sleeping on it I thought you were right and `firstExistingFile` was better :)

Thanks for the help!

Issue #42, opened 2016-11-22
Show less noisy error message when no changelog commits are found


lekkas:
Currently running `versionist` when no changelog commits are made produces a stacktrace:

```
No commits to calculate the next increment level from
Error: No commits to calculate the next increment level from
    at Object.exports.calculateNextIncrementLevel (/usr/local/lib/node_modules/versionist/lib/semver.js:133:11)
    at Object.exports.calculateNextVersion (/usr/local/lib/node_modules/versionist/lib/versionist.js:163:33)
    at async.waterfall (/usr/local/lib/node_modules/versionist/bin/cli.js:145:32)
    at nextTask (/usr/local/lib/node_modules/versionist/node_modules/async/dist/async.js:5208:14)
    at /usr/local/lib/node_modules/versionist/node_modules/async/dist/async.js:5202:13
    at apply (/usr/local/lib/node_modules/versionist/node_modules/async/dist/async.js:41:25)
    at /usr/local/lib/node_modules/versionist/node_modules/async/dist/async.js:76:12
    at /usr/local/lib/node_modules/versionist/node_modules/async/dist/async.js:988:16
    at versionist.readCommitHistory (/usr/local/lib/node_modules/versionist/bin/cli.js:140:14)
    at ChildProcess.child.on (/usr/local/lib/node_modules/versionist/lib/versionist.js:108:12)
```

PR #63, opened 2016-12-20
Only output stack trace if DEBUG=versionist (#42)

craig-mulligan:
connects to #42
change-type: patch

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -32,10 +32,11 @@ const versionist = require('../lib/versionist');
 const semver = require('../lib/semver');
 const configuration = require('../lib/cli/configuration');
 const packageJSON = require('../package.json');
+const debug = require('debug')(packageJSON.name);
 
 const showErrorAndQuit = (error) => {
   console.error(chalk.red(error.message));
-  console.error(chalk.red(error.stack));
+  debug(chalk.red(error.stack));
   console.error('Join our Gitter channel if you need any help!');
   console.error('  https://gitter.im/resin-io/versionist');
   process.exit(1);
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   "dependencies": {
     "async": "^2.0.0-rc.6",
     "chalk": "^1.1.3",
+    "debug": "2.5.1",
     "handlebars": "^4.0.5",
     "handlebars-helpers": "^0.6.1",
     "js-yaml": "^3.6.1",


lurch:
This also makes the unit-tests' output much nicer :)

Maybe I'm too much of a Linux-head (and / or maybe I'm not aware of existing NodeJS conventions), but would it make sense to change this to use:
```javascript
  if (process.env.DEBUG === 'true' || process.env.DEBUG === '1') {
```
instead?

Page-:
It might be worth using a module like https://www.npmjs.com/package/debug

craig-mulligan:
@lurch As far as I've seen node convention seems to be `true` rather than `1` but non-node users will use this so we should catch all. 

@Page- All errors are thrown up to `showErrorAndQuit` handler so pretty easy to manage right now. But I agree it would be useful to have levels etc. @lekkas @lurch What do you think?

lurch:
:+1: to supporting both `true` and `1`, but I have no opinion either way on different debug levels. @jviotti ?

lekkas:
I'd be fine with using the `debug` module, which btw doesn't support levels, it will just require a `DEBUG=versionist` (or `DEBUG=*`) convention to print logs in stderr with `console.error`

I'm also totally fine with keeping the current solution which checks directly the `DEBUG` variable, but I'd prefer if we only checked if it's set, rather than comparing with either `true` or `1` (which is what we also use in resin-cli https://github.com/resin-io/resin-cli/blob/master/lib/errors.coffee#L25)

jviotti:
I like the proposal of using the `debug` model and make use of different levels (you can do this by having a convention such as `DEBUG=versionist:<level>`) or use any other similar module.

lekkas:
I think it's an overkill to add debug levels as part of this PR, maybe we can in the future, by first creating a new issue where we can discuss on a way to do it.

jviotti:
Yeah, I agree. The current convention, as far as I've seen, is setting DEBUG to the name of the program (e.g: versionist). `debug` does this out of the box, so I guess we can make a quick, simple integration now, and make use of its more advanced features when we see fit.

craig-mulligan:
@jviotti could you take another look when you are back.

jviotti:
Looks good, I'll approve. The PR title should be changed before merging though, since it still refers to `DEBUG=true`.

jviotti:
Ugh, for some reason I can't approve from mobile :/ I'll do it as soon as I get access to a computer, otherwise consider it approved :)

PR #72, opened 2017-02-08
Show nice error message instead of crashing if version incrementing commits are missing

### LINKED ISSUE #65: Error: Cannot read property 'trim' of null
Attempting to run `versionist` without any arguments in the `versinist` repo (commit 7862dd5) produces the following error:

```
Cannot read property 'trim' of null
TypeError: Cannot read property 'trim' of null
    at Function.clean (/usr/local/lib/node_modules/versionist/node_modules/semver/semver.js:266:24)
    at Object.module.exports.updateVersion.npm (/usr/local/lib/node_modules/versionist/lib/presets.js:271:37)
    at apply (/usr/local/lib/node_modules/versionist/node_modules/lodash/lodash.js:499:17)
    at Object.wrapper [as updateVersion] (/usr/local/lib/node_modules/versionist/node_modules/lodash/lodash.js:5383:16)
    at async.waterfall (/usr/local/lib/node_modules/versionist/bin/cli.js:178:26)
    at nextTask (/usr/local/lib/node_modules/versionist/node_modules/async/dist/async.js:5070:14)
    at /usr/local/lib/node_modules/versionist/node_modules/async/dist/async.js:5064:13
    at apply (/usr/local/lib/node_modules/versionist/node_modules/async/dist/async.js:21:25)
    at /usr/local/lib/node_modules/versionist/node_modules/async/dist/async.js:56:12
    at /usr/local/lib/node_modules/versionist/node_modules/async/dist/async.js:840:16
```

This looks like it could be related to that commit missing a change-type entry, but still, it'd be nice if we handled this error case more gracefully

lekkas:
Closes #65

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -153,8 +153,7 @@ async.waterfall([
     });
 
     if (_.includes(documentedVersions, version)) {
-      console.log(`Omitting: ${version}`);
-      return callback(null, null);
+      return callback(new Error(`No commits were annotated with a change type since version ${version}`), null);
     }
 
     const entry = versionist.generateChangelog(history, {
--- a/tests/cli/cases/missing-changetype-commits.js
+++ b/tests/cli/cases/missing-changetype-commits.js
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2016 Resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const m = require('mochainon');
+const shelljs = require('shelljs');
+const utils = require('../utils');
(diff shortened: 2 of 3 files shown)


bin/cli.js
```diff
@@ -154,7 +154,7 @@ async.waterfall([
 
     if (_.includes(documentedVersions, version)) {
       console.log(`Omitting: ${version}`);
```
craig-mulligan: Now that we have the debug package should we start moving `console.logs` to it? Or best to do it all at once?

lekkas: Good catch! Btw do we even need this 'Omitting' message here? if no versions are found `versionist` will exit anyway with the message that describes what went wrong.

lekkas:
Added tests

tests/cli/cases/missing-changetype-commits.js
```diff
@@ -0,0 +1,62 @@
...
+  '  },',
+  '  template: [',
+  '    \'## {{version}}\',',
+  '    \'\',',
+  '    \'{{#each commits}}\',',
+  '    \'{{#with footer}}\',',
+  '    \'- {{capitalize Changelog-Entry}}\',',
+  '    \'{{/with}}\',',
+  '    \'{{/each}}\'',
+  '  ].join(\'\\n\')',
```
lurch: Yuck! ;-)  I guess this would be more readable if it was simply read in from an external file?

lekkas: I've followed @jviotti approach from the existing test cases in `tests/cli/cases/` .

We can definitely add a separate folder to keep external files (e.g. `fixtures/`) but if we want to refactor tests I'd vote for opening a separate issue/PR.

bin/cli.js
craig-mulligan: ah I see, yea we don't need that.

PR #77, opened 2017-03-09
Update README.md

alisondavis17:
Change support section to point to Forums instead of Gitter

### DIFF
--- a/README.md
+++ b/README.md
@@ -607,7 +607,7 @@ Support
 -------
 
 If you're having any problem, please [raise an issue][github-issue] on GitHub,
-or join our [Gitter channel][gitter] and the Resin.io team will be happy to
+or find us on the [forums][https://forums.resin.io], and the resin.io team will be happy to
 help.
 
 License


lurch:
If we're pointing people at our forums instead of Gitter, should that be accompanied by the creation of a `versionist`-specific forum?

alisondavis17:
Sounds like a good question for @imrehg!

We also don't have to point to the forums at all. We could simply remove Gitter and keep the Support section focused on GitHub issues. The main goal is to remove the expectation of Gitter as a primary Support channel.

craig-mulligan:
I'd don't think it matters too much as long as we have consistency with regards to it all open source projects. @alisondavis17

I suggest we keep it to just gh-issues until the project is more mature? @alisondavis17

lurch:
> as long as we have consistency with regards to it all open source projects. 

Does that mean something similar should also be done for Etcher?

Ping @jviotti and @alexandrosm

lurch:
I just noticed that `versionist` itself is recommending Gitter too when it shows any error messages:
https://github.com/resin-io/versionist/blob/master/bin/cli.js#L40

imrehg:
Closing this and continuing in #98

Issue #80, opened 2017-04-18
Make change-type case-insensitive


craig-mulligan:
```
resin-site git:(master) ✗ versionist -d
Invalid increment level: Patch
Join our Gitter channel if you need any help!
  https://gitter.im/resin-io/versionist
```
let's make it case insensitive?

Page-:
I just hit this as well with `Minor`

jhermsmeier:
I've just run into this on v2.11.2:

```
[18:48:58] ~/W/etcher (release-1.2.0|*2) $ versionist
Invalid increment level: Patch
Join our Gitter channel if you need any help!
  https://gitter.im/resin-io/versionist

[18:49:06] ~/W/etcher (release-1.2.0|*2) $ versionist -v
2.11.2
```

jhermsmeier:
Nevermind; custom versionist.conf was at fault :/

Issue #76, opened 2017-03-08
'None' change-type suggestion


lurch:
As discussed in https://github.com/resin-io-modules/resin-procbots/issues/16 it might be useful if `versionist` supported a `change-type: none` footer tag, so that you could indicate that a commit is only doing small "typo" fixes (which don't modify the program behaviour e.g. editing a code-comment), and such a commit should neither bump the version nor add any entries to the Changelog.
Otherwise there's a risk that the Changelog may get polluted with "typo fix" entries, which might dissuade people from bothering to fix typos on the first place?

CameronDiver:
I was literally just about to make this issue myself! I made an inconsequential change to a `package.json`, which doesn't need a change-type at all, but with VersionBot I needed to add one, so...

+1

PR #81, opened 2017-04-20
Make change-type insensitive & add change-type none

craig-mulligan:
Connects to #80 and #76

There are two ways to add the `change-type: none` flag, internally we map the 3 levels to the semver lib so we could add a fourth level 'none' and then drop it before passing to semver but this was a far simpler approach.

Change-Type: patch

### DIFF
--- a/versionist.conf.js
+++ b/versionist.conf.js
@@ -1,8 +1,16 @@
-var execSync = require('child_process').execSync;
+'use strict';
 
-var getAuthor = (commitHash) => {
-  return execSync(`git show --quiet --format="%an" ${commitHash}`, { encoding: 'utf8' }).replace('\n', '');
-}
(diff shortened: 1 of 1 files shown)


lurch:
Should this result in a version-bump and new release of versionist?

craig-mulligan:
@lurch it's available now on `v2.9.0`.

jhermsmeier:
I've just run into this on v2.11.2:

```
[18:48:58] ~/W/etcher (release-1.2.0|*2) $ versionist
Invalid increment level: Patch
Join our Gitter channel if you need any help!
  https://gitter.im/resin-io/versionist

[18:49:06] ~/W/etcher (release-1.2.0|*2) $ versionist -v
2.11.2
```

Issue #85, opened 2017-05-31
First class Rust Cargo support


majorz:
Currently Versionist supports NPM's package.json files. The Rust counterpart of NPM packages are Cargo crates and the counterpart of package.json is Cargo.toml. Since those are very equivalent scenarios Versionist can have a first-class support for Rust/Cargo.

lurch:
See also #84 ;-)

PR #86, opened 2017-05-31
First class Rust Cargo support

majorz:
Here is an equivalent Cargo support to the NPM support we have. It replaces the version string in the Cargo.toml file located in the root of each Rust project. The easiest solution for me was to do a search and replace by regex using the `replace-in-file` NPM package.

**_Please, NOTE that I have tested the added exact JS snippet only locally with Node without running Versionist. I have not added any tests as well._**

Connects-To: #85


lib/presets.js
```diff
@@ -283,6 +284,43 @@ module.exports = {
...
+      const cargoToml = path.join(cwd, 'Cargo.toml');
+      const cleanedVersion = semver.clean(version);
+
+      if (!cleanedVersion) {
+        return callback(new Error(`Invalid version: ${version}`));
+      }
+
+      replace({
+        files: cargoToml,
+        from: /(version\s*=\s*)"(?:0|[1-9]\d*)\.(?:0|[1-9]\d*).(?:0|[1-9]\d*)"/g,
```
hedss: `.` between the second and third digit capture should be `\.`.

lib/presets.js
```diff
@@ -283,6 +284,43 @@ module.exports = {
...
+        to: '$1"' + cleanedVersion + '"'
```
hedss: This could be `$1"${cleanedVersion}"` to use the same format as the other version methods.

majorz: Unfortunately it does not support template literals and outputs "${cleanedVersion}" directly to the file.

majorz:
I tested it with Versionist cli.js and it works well :icecream:

package.json
```diff
@@ -39,6 +39,7 @@
     "semver": "^5.2.0",
     "touch": "^1.0.0",
     "update-json": "^1.0.0",
-    "yargs": "^4.7.1"
+    "yargs": "^4.7.1",
+    "replace-in-file": "^2.5.0"
```
CameronDiver: Could you add this with `npm install --save replace-in-file` please? That way it will be added alphabetically. Alternatively you can just move it into the correct position yourself, but I thought I should let you know about `npm i --save` :)

majorz: Thanks for the tip. I run the command and additionally it added a new line at the end of the file.

Issue #96, opened 2017-11-01
Update Cargo.lock if exists side by side with Cargo.toml


majorz:
A Rust binary executable project will have Cargo.lock side by side with Cargo.toml: http://doc.crates.io/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

We need to modify Cargo.lock in the same way we modify Cargo.toml to support properly Rust binary executable projects, e.g. WiFi Connect.

lurch:
Closing this, as it was added by #97

PR #97, opened 2017-11-01
Update Cargo.lock if exists side by side with Cargo.toml

majorz:
Change-Type: patch
Connects-To: #96
---- Autogenerated Waffleboard Connection: Connects to #96


lib/presets.js
```diff
@@ -305,15 +305,15 @@ module.exports = {
      * });
      */
     cargo: (options, cwd, version, callback) => {
-      const cargoToml = path.join(cwd, 'Cargo.toml');
+      const cargoFiles = path.join(cwd, 'Cargo.*');
```
lurch: Do we really want such a wide-reaching glob, that would also match e.g. `Cargo.readme`?
(I've never done any Rust stuff, so I dunno if that's a silly question or not :wink: )

majorz: It is only those two that are used for Rust projects. If someone puts a Cargo.xyz in the root, there still should be something inside which matches the regex. But people are doing strange things by nature and there really exists a chance for this by the universal laws. Although this conversation probably diminished a big part of it :D

PR #98, opened 2017-12-07
update support links and info

imrehg:
Superseding #77, based on that conversation update support CTA to GitHub Issues.

Outstanding question: should the Gitter badge be removed from the ReadMe? Or keep it there anyways. It is kept for the moment.

Let me know if any of the text should be rewritten.


alisondavis17:
@imrehg nice catch. In general, we should have this match the support section of the README in our other repos - such as https://github.com/resin-io/resin-cli

imrehg:
@alisondavis17 so you mean, that don't have a separate gitter channel for this (as it is now), but point to the resin-io gitter (as it is in resin-cli)? And/or the description in the `Support` section should match between projects (I see that the other one has the same approach as this, just not the same words, can fix that)

lurch:
Hmmm, https://github.com/resin-io/resin-cli/blob/master/README.md#support doesn't mention gitter at all, but there _is_ still a "gitter: join chat" button at the top?
Looping in @pimterry too.

lurch:
And should https://gitter.im/resin-io/versionist be "switched off" ?

imrehg:
@lurch was thinking the same way (last message is by us in September) and changed to the resin-io chat here too.

@alisondavis17 rewritten, taking the cue from resin-cli, let me know if there are any language issues!

cc @jviotti

alisondavis17:
lgtm @imrehg thanks!

imrehg:
@lurch what do you think?

Issue #99, opened 2018-01-18
Python __init__.py support


nghiant2710:
In the Python SDK, we use set the package version in `__init__.py` so need to add support for that file the same as NPM's package.json or Rust Cargo.toml

Edited by @lurch to add the back-ticks around `__init__.py` :wink:

lurch:
Closing this, as #100 added this feature

PR #100, opened 2018-01-18
Python __init__ file support

nghiant2710:
Connects-to: #99
Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #99

### DIFF
--- a/lib/presets.js
+++ b/lib/presets.js
@@ -321,8 +321,49 @@ module.exports = {
           return callback(error);
         }
       });
-    }
+    },
 
+    /**
+     * @summary Update package version in Python Init file
+     * @function
+     * @public
+     *
+     * @param {Object} options - options
+     * @param {String} [options.targetFile] - path to target python file, defaults to `__init__.py`
+     * @param {String} cwd - current working directory
+     * @param {String} version - version
+     * @param {Function} callback - callback (error)
+     * @returns {null}
+     *
+     * @example
+     * presets.updateVersion.initPy({}, process.cwd(), '1.0.0', (error) => {
+     *   if (error) {
+     *     throw error;
+     *   }
+     * });
+     */
+    initPy: (options, cwd, version, callback) => {
+      _.defaults(options, {
+        targetFile: '__init__.py'
+      });
+
+      const initFile = path.join(cwd, options.targetFile);
+      const cleanedVersion = semver.clean(version);
+
+      if (!cleanedVersion) {
+        return callback(new Error(`Invalid version: ${version}`));
+      }
+
+      replace({
+        files: initFile,
+        from: /(__version__\s*=\s*)('|")(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\2/g,
+        to: '$1$2' + cleanedVersion + '$2'
+      }, (error) => {
+        if (error) {
+          return callback(error);
+        }
+      });
+    }
   },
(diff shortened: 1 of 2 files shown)


lib/presets.js
```diff
@@ -321,8 +321,48 @@ module.exports = {
...
       });
-    }
+    },
+
+    /**
+     * @summary Update package version in Python Init file
+     * @function
+     * @public
+     *
+     * @param {Object} options - options
```
lurch: Is it worth documenting the `options.targetFile` here?
I assume on `resin-sdk-python` you'll need to pass `{targetFile: 'resin/__init__.py'}` ?

lib/presets.js
```diff
@@ -321,8 +321,48 @@ module.exports = {
...
+      const cleanedVersion = semver.clean(version);
+
+      if (!cleanedVersion) {
+        return callback(new Error(`Invalid version: ${version}`));
+      }
+
+      replace({
+        files: initFile,
+        from: /(__version__\s*=\s*)'(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)'/g,
+        to: '$1\'' + cleanedVersion + '\''
```
lurch: I guess this would break if somebody happened to have `__version__ = "1.0.0"` in their `__init__.py` file? You might be able to allow for both styles of quotes with something like:
```
from: /(__version__\s*=\s*)('|")(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\2/g,
to: '$1$2' + cleanedVersion + '$2'
```
?

lib/presets.js
nghiant2710: I've updated the docs and yes, we need to pass `resin/__init__.py` for the Python SDK https://github.com/resin-io/resin-sdk-python/pull/63.

lib/presets.js
nghiant2710: @lurch nice, I've not noticed this 👍

lurch: @majorz I know nothing about Rust or Cargo files, but do they allow both double-quotes and single-quotes? If they do, perhaps a similar fix to what I've suggested here would also be useful to the Rust support you added to versionist?

lurch:
Could you also modify this PR to update https://github.com/resin-io/versionist#updateversion as appropriate? :)

nghiant2710:
@hedss done!

lib/presets.js
majorz: @lurch I checked the TOML specification and there are single-quote strings as well - literal strings, which cannot be escaped. I will PR this.

lurch:
@nghiant2710 I think you should have merged this using the `procbots/versionbot/ready-to-merge` label, rather than merging it manually?
@hedss Should manual PR merging be disabled for this repo?

nghiant2710:
@hedss @lurch sorry, my bad :/ do we need to revert this PR?

lurch:
Nah, I think VersionBot will sort everything out automatically the next time that the `procbots/versionbot/ready-to-merge` label _is_ used to merge a PR?

hedss:
@lurch It will, yep. And I've just disabled manual merging. The only problem is, something else needs to be deployed to versionist first. I'm going to make the updates manually for now and publish to NPM.

hedss:
@lurch Ah! I see that you PRd something afterwards, excellent!

Issue #101, opened 2018-01-22
Update the version string of the Rust package, not the dependencies


majorz:
There can be multiple `version = "..."` entries in both Cargo.toml and Cargo.lock - for the main package and for dependencies as well. We need to change the version string of the package only.

Also update the regex expression for Rust manifest files to support version numbers wrapped in single-quote literal strings.

And narrow the regex replacement to just the .toml and .lock files (not Cargo.*).

majorz:
Oops, there can be multiple `version = "..."` entries in both files - for the main package and for dependencies as well. I will change the title and description of the issue.

lurch:
> And narrow the regex replacement to just the .toml and .lock files (not Cargo.*).

Heh ;-)  https://github.com/resin-io/versionist/pull/97#discussion_r148371165

Hmm, from a little bit of searching, it seems like `Cargo.lock` is broadly similar to `npm-shrinnkwrap.json`.
@jviotti @jhermsmeier @hedss @lekkas  On projects (like Etcher) where we actually use both a `package.json` **and** a `npm-shrinkwrap.json`, and where VersionBot is being used, I guess we need an additional `updateVersion` [preset](https://github.com/resin-io/versionist#updateversion) that would update the version number in both `package.json` and `npm-shrinkwrap.json` at the same time? If this _is_ the case (and I'm not just getting horribly confused?) I'll create a separate `versionist` issue to track this.

majorz:
@lurch +1 for updating `npm-shrinkwrap.json` if exists in the `npm` preset. This should be done for `package-lock.json` as well I think.

lurch:
Cool, created #103 :)

lurch:
This can be closed now, @majorz ?

majorz:
Right, closing it now, thanks!

PR #104, opened 2018-01-22
Update the proper version definition of Rust Cargo manifests

majorz:
We are doing more heavy regex processing now.

First the package name is extracted from Cargo.toml. It is located in the  `[package]` section of Cargo.toml. The package name is used for detecting the proper `version = "..."` definition in Cargo.lock. Cargo.lock is generated automatically, so `version` always follows `name`, and the code takes advantage of that fact for simplification purposes.

The proper `version = "..."` definition in Cargo.toml is inside the `[package]` section, so the regex looks for first occurrence after `[package]`.

Change-Type: patch
Connects-To: #101
---- Autogenerated Waffleboard Connection: Connects to #101


majorz:
I have not linted the code... Will fix and recommit now.

lib/presets.js
```diff
@@ -305,21 +305,59 @@ module.exports = {
...
+          files: cargoToml,
+          from: /(\[package\](?:\n|.)*?version\s*=\s*)("|')(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)("|')/m,
+          to: '$1$2' + cleanedVersion + '$3'
+        }, (err) => {
+          if (err) {
+            return callback(err);
+          }
+        });
+
+        // Update first `version = "..."` occurrence immediately after `name = "${packageName}"`
```
lurch: Should this whole chunk (and also the bit that extracts the packageName) be inside a `if (fs.existsSync(cargoLock))` ?

lib/presets.js
```diff
@@ -305,21 +305,59 @@ module.exports = {
...
+          return callback(
+            new Error(`Package name not found in ${cargoToml}`)
+          );
+        }
+        const packageName = matches[1];
+
+        // Update first `version = "..."` occurrence immediately after `[package]`
+        replace({
+          files: cargoToml,
+          from: /(\[package\](?:\n|.)*?version\s*=\s*)("|')(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)\.(?:0|[1-9]\d*)("|')/m,
```
lurch: Minor tweak, but the second `("|')` could/should be replaced by `\2`, as the current regex will also match `version = "1.0.0'` or `version = '1.0.0"`, whereas using `\2` will ensure that the quotes match up (i.e. it'll only match `version = "1.0.0"` or `version = '1.0.0'`).

lib/presets.js
majorz: Yes, it should. Nice catch!

majorz:
I am going to add tests for all of this, since it is getting quite complex and there should be tests covering the `cargo` preset anyway.

lib/presets.js
majorz: Excellent

lib/presets.js
```diff
@@ -305,21 +305,59 @@ module.exports = {
...
         if (error) {
+          if (error.code === 'ENOENT') {
+            error.message = `No such file or directory: ${cargoToml}`;
+          }
+
           return callback(error);
         }
+
+        // Capture first `name = "..."` occurrence immediately after `[package]`
+        const matches = contents.match(/\[package\](?:\n|.)*?name\s*=\s*(?:"|')(.+)(?:"|')/m);
```
lurch: I did a bit of searching, and it [looks like](https://stackoverflow.com/a/47317639) `[^]` might be a better version of `(?:\n|.)` ?

majorz: Nice, that will handle `\r` as well.

lurch:
> it is getting quite complex

I'm not familiar with constructing & combining regexes in JS, but perhaps you could also simplify it by doing  something like:
```js
semverDigitRe = '(?:0|[1-9]\d*)'
semverRe = `${semverDigitRe}\.${semverDigitRe}\.${semverDigitRe}`
// and then use semverRe in the other regexes...
```
?

majorz:
Actually the semver regex is incomplete, since it matches `1.2.3`, but not `1.2.3-alpha.10`. I will see what options are available for fixing this as well.

lurch:
https://github.com/resin-io-modules/resin-semver might be a good starting point? (i.e. have code you can steal)

majorz:
Hah, I was just looking at this. We only need the semver regex for validation purposes, but maybe it is not our job to do this and we can just pass through the content between the quotes?

lurch:
> maybe it is not our job to do this and we can just pass through the content between the quotes?

Hmm, good point. In theory at least, versionist could also be used for non-semver [versioning schemes](https://github.com/resin-io/versionist#incrementversion-functionstring), so we probably shouldn't be assuming that the old version-string is actually a semver?

majorz:
I just checked how Cargo parses the version input and it requires it to be a semver, but is more strict than the regexes from the JS libraries we are looking into, e.g. it won't allow a `v` prefix.

Let's say a user has an incorrect version string - he has not verified that it is working with a `cargo build` or `cargo run`. I think versionist should still update the version string to the one incremented from the README file. Let's leave the validation out.

majorz:
I included tests and I also refactored the code to use async.waterfall, since during writing the tests I found out that the callback was not called on success.

lib/presets.js
```diff
@@ -305,22 +305,55 @@ module.exports = {
...
-        if (error) {
-          return callback(error);
+      async.waterfall([
+        (done) => {
+          return fs.readFile(cargoToml, 'utf8', done);
+        },
+
+        (contents, done) => {
+          // Capture first `name = "..."` occurrence immediately after `[package]`
+          const matches = contents.match(/\[package\][^]+?name\s*=\s*("|')(.+)(\1)/m);
```
lurch: nit: I think there's no need for brackets around the last `\1` ?
And the `(.+)` should be `(.+?)` ?

lib/presets.js
```diff
@@ -305,22 +305,55 @@ module.exports = {
...
+          } else {
+            done(null, matches[2]);
+          }
+        },
+
+        (packageName, done) => {
+          if (fs.existsSync(cargoLock)) {
+            // Update first `version = "..."` occurrence immediately after `name = "${packageName}"`
+            const cargoLockRe = new RegExp(
+              `(name\\s*=\\s*(?:"|')${packageName}(?:"|')[^]+?version\\s*=\\s*)("|').*(?:\\2)`, 'm'
```
lurch: nit: I think `(?:\\2)` can just be replaced by `\2`?
And I suspect you definitely want to replace the `.*` with `.*?`

lib/presets.js
```diff
@@ -305,22 +305,55 @@ module.exports = {
...
+          } else {
+            done(null, null);
+          }
+        },
+
+        (_changedFiles, done) => {
+          // Update first `version = "..."` occurrence immediately after `[package]`
+          replace({
+            files: cargoToml,
+            from: /(\[package\][^]+?version\s*=\s*)("|').*(?:\2)/m,
```
lurch: As above, `.*` -> `.*?` and `(?:\2)` -> `\2` ?

tests/presets.spec.js
```diff
@@ -740,6 +740,224 @@ describe('Presets', function() {
...
+            ''
+          ].join('\n'));
+        });
+
+        afterEach(function() {
+          fs.unlinkSync(this.cargoToml);
+          this.cwd.removeCallback();
+        });
+
+        it('should be able to update the version', function(done) {
```
lurch: Is the description here wrong? Should it say something like `it('should yield an error', ` ?

majorz:
@lurch, thanks a lot for pointing out those, I fixed them all now!

lurch:
Looking great! :-)
Could you also update the README to say e.g. "This preset updates the `version` property of `$CWD/Cargo.toml` (and also `$CWD/Cargo.lock` if it exists)."

Given that Cargo.lock is autogenerated, I suppose we'll never run into the situation of the `name` from Cargo.toml **not** appearing in Cargo.lock? (is it worth adding a test-case for that?)

majorz:
> Given that Cargo.lock is autogenerated, I suppose we'll never run into the situation of the `name` from Cargo.toml **not** appearing in Cargo.lock? (is it worth adding a test-case for that?)

The nice thing about adding similar test is not only to ensure correctness in this situation, but also to ensure overall correctness and stability. Those corner cases may unveil other issues, so +1 from me. I will add it now.

majorz:
It turned out that the `replace` function does not produce an error when the target file is not updated - it passes an empty list as a result (changedFiles is []). This is true for any malformed Cargo.toml/Cargo.lock that does not match the regular expressions. I added handling for that in the async waterfall and also two test cases targeting both files.

At this point we need to add corresponding changes to the Python handler as well:
https://github.com/resin-io/versionist/blob/900db381dfcdbd69024ab67aed41766553a7f843/lib/presets.js#L357

Tomorrow I am going to create a new issue for it, add fixes and tests and then create a PR.

lib/presets.js
```diff
@@ -305,22 +305,67 @@ module.exports = {
...
+          // Update first `version = "..."` occurrence immediately after `[package]`
+          replace({
+            files: cargoToml,
+            from: /(\[package\][^]+?version\s*=\s*)("|').*?\2/m,
+            to: '$1$2' + cleanedVersion + '$2'
+          }, done);
+        },
+
+        (changedFiles, done) => {
+          if (_.isArray(changedFiles) && _.isEmpty(changedFiles)) {
```
lurch: It obviously doesn't hurt anything, but I guess _strictly_ speaking the `_.isArray` isn't needed here?

tests/presets.spec.js
```diff
@@ -740,6 +740,289 @@ describe('Presets', function() {
...
+          this.cwd = tmp.dirSync();
+        });
+
+        afterEach(function() {
+          this.cwd.removeCallback();
+        });
+
+        it('should yield an error', function(done) {
+          presets.updateVersion.cargo({}, this.cwd.name, '1.0.0', (error) => {
+            m.chai.expect(error).to.be.an.instanceof(Error);
```
lurch: Now that there's errors being emitted from several places in the code, is it also worth checking the error-message in these unit-tests to check that the right error is being thrown by the right test? \*shrug\*

tests/presets.spec.js
```diff
@@ -740,6 +740,289 @@ describe('Presets', function() {
...
+          ].join('\n'));
+        });
+
+        afterEach(function() {
+          fs.unlinkSync(this.cargoToml);
+          fs.unlinkSync(this.cargoLock);
+          this.cwd.removeCallback();
+        });
+
+        it('should be able to update the version', function(done) {
```
lurch: Just a small nit, but is it worth updating this test description to something like `it('should be able to update the version in both files', ` ?

tests/presets.spec.js
```diff
@@ -740,6 +740,289 @@ describe('Presets', function() {
...
+            ''
+          ].join('\n'));
+        });
+
+        afterEach(function() {
+          fs.unlinkSync(this.cargoToml);
+          this.cwd.removeCallback();
+        });
+
+        it('should be able to update the version', function(done) {
```
lurch: Perhaps this description could be expanded to `it('should be able to update the version (and preserve the single-quotes)', ` ?

lurch:
Couple of other thoughts I've had:
* I suppose there's not really much point in adding a (failing) test for having a `Cargo.lock` but no `Cargo.toml` ?
* Probably also little point in checking that updating to the same version means that the updated file is identical to the original file?
* Given the regex-improvements you made to explicitly _not_ look for a semver-version, is it worth adding some test-cases for that? (i.e. check that `version = "nonsense"` gets successfully updated to `version = "1.1.0"`)
* Would this (invalid) Cargo.lock pass the tests?
```
[[package]]
name = "foo"
dependencies = [
 "bar 2.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
]

[metadata]
"checksum bar 2.0.0 (registry+https://...)" = "..."

[[package]]
name = "bar"
version = "2.0.0"
source = "registry+https://github.com/rust-lang/crates.io-index"

```
However this is so contrived that it might not be worth adding the effort to 'detect' files malformed like this.

> At this point we need to add corresponding changes to the Python handler as well:
Tomorrow I am going to create a new issue for it, add fixes and tests and then create a PR.

Fantastic! :grinning: I was going to suggest that to @nghiant2710 after this PR got merged ;)

P.S. I apologise if I'm coming across as overly picky, but feel free to ignore any of my suggestions in this comment.

lib/presets.js
majorz: In the case above this one `_.isArray` is needed, because we want to ensure that the passed value is an empty array (`_.isEmpty(null)` is true in the case above). I placed it here as well for consistency. It is not needed here though in terms of correctness as you mention.

The problem is that in the first case `changedFiles` can be both a `null` and an array. In the second: an array only.

Maybe it is better to change `_.isArray(changedFiles)` to `!_.isNull(changedFiles)`, which describes better what is happening.

What's your opinion on all of this?

tests/presets.spec.js
majorz: Maybe there is an error-type kind of thing that can be used - I will check and let you know.

tests/presets.spec.js
majorz: +1

tests/presets.spec.js
majorz: +1

majorz:
> * I suppose there's not really much point in adding a (failing) test for having a `Cargo.lock` but no `Cargo.toml` ?
I will add such a test for completeness.

> * Probably also little point in checking that updating to the same version means that the updated file is identical to the original file?
I will add a single one for such a case.

> * Given the regex-improvements you made to explicitly _not_ look for a semver-version, is it worth adding some test-cases for that? (i.e. check that `version = "nonsense"` gets successfully updated to `version = "1.1.0"`)
I will add such a test - in case in the future somebody adds such a semver regex. Our intention is to allow input like `version = "nonsense"`.

> * Would this (invalid) Cargo.lock pass the tests?
> ...
> However this is so contrived that it might not be worth adding the effort to 'detect' files malformed like this.
Such invalid Cargo.lock will be treated incorrectly indeed. Now that I am thinking about this again, it can be fixed if we extend somehow `[^]+?` to reject `[`. I will check how this can be done. 

> P.S. I apologise if I'm coming across as overly picky, but feel free to ignore any of my suggestions in this comment.
Not coming this way.

tests/presets.spec.js
majorz: `Error` has a `name` property that looks like the natural way to determine error types. Another way is to use custom error types, but that is too much for this case I think.

lib/presets.js
lurch: I'm happy for it to be left as-is ("doesn't hurt anything"), I was merely asking to check that I understood exactly what the code was doing ("In the second: an array only"), and you've confirmed that I have :-)

tests/presets.spec.js
lurch: Ahh, if there's no easy way to get the actual error-message during the unit-tests, then I'm happy with what you've got currently.

majorz: Hm, maybe we can check the actual message with string formatting only...

lurch:
@majorz I mentioned in today's Etcher meeting that I was helping you with your regexes on this, and @jhermsmeier said you definitely shouldn't be using regexes to parse complex files like this, he pointed me at https://github.com/BinaryMuse/toml-node
:-)

majorz:
@lurch, this one is a parser only. We need to serialize back the changed content.

There was not any available npm package, which could parse a Toml file and write back the changes by preserving the rest of the file intact. For example tomlify-j0.4 inserted commas at places the original did not had. This is important because when versionist changes the files, the Rust compiler will modify Cargo.lock in a way that git will consider a changed file.

We can exclude the option of writing such Toml serialization on our own. The only other way I can think of is doing what the regexes are doing with custom code, but that seems unnecessary.

I could not implement the changes above today as I was on afternoon support shift, so I will continue on Monday.

majorz:
I should have added that I would not suggest either regexes for complex parsing and that I think they fit the heuristic approach of this case only. I am definitely open for better ideas though (e.g. for future improvements).

lurch:
ping @jhermsmeier in case he has any good ideas.

jhermsmeier:
@majorz @lurch unfortunately not – my recommendation was premature. I've tested some of the TOML libs now, and none of them parse / serialize what we'd need properly... we'd have to write our own, or fork and improve existing ones – so regular expressions seem like the way for now.

lurch:
> my recommendation was premature

IMHO it's always worth the time investigating other options, even if you end up rejecting them :)

lib/presets.js
majorz: Leaving it as is.

tests/presets.spec.js
majorz: String formatting worked..

majorz:
> * Probably also little point in checking that updating to the same version means that the updated file is identical to the original file?

@lurch With this case I hit an issue. The `replace` function from `replace-in-file` returns an empty array when the regular expression matches the content, but the final content is equal to the initial one. That is when we want to update using the version already in the file.

However an empty array is returned when the regular expression does not match the contents of the file, which we treat as an update error.

I am thinking of using the code from `replace-in-file` for writing a simplified version of the `replace` function, which is suitable for our use case only. The original `replace` accepts replacement in multiple files, etc., but this is not really needed as I originally used it, since our application logic requires updating files explicitly one by one.

A simpler `replace` tuned for our own use case will simplify the application logic as well, since there will no longer be an array of file names to be returned on success.

What is your opinion on all of that?

lurch:
> The replace function from replace-in-file returns an empty array when the regular expression matches the content, but the final content is equal to the initial one. That is when we want to update using the version already in the file.
However an empty array is returned when the regular expression does not match the contents of the file, which we treat as an update error.

Doh, how annoying! Maybe it's also worth filing a bug / feature-request against the `replace-in-file` module?
I feel less silly for suggesting such a simple test-case now ;-)

> I am thinking of using the code from `replace-in-file` for writing a simplified version of the `replace` function, which is suitable for our use case only.

Sounds good to me, and I guess this will also be useful for the `initPy` preset.
Pinging @jhermsmeier in case he has any good ideas or suggestions.

Hmmm, I guess an alternative way of doing it might be to use the regex to actually check what version is currently stored in the file, and then skip calling replace-in-file if the version is already the same?

majorz:
> Doh, how annoying! Maybe it's also worth filing a bug / feature-request against the `replace-in-file` module?

I checked the `replace-in-file` sources and the code on their side is purposefully implemented like that, so I am not sure they will agree on extending/modifying the behavior.

> Hmmm, I guess an alternative way of doing it might be to use the regex to actually check what version is currently stored in the file, and then skip calling replace-in-file if the version is already the same?

I initially thought about that as well, but having additional hairy regex for bypassing the `replace-in-file` shortcoming led me to think about implementing `replace` ourselves :D

lurch:
> additional hairy regex

Oh, I was assuming it'd be easy to re-use the regexes we already have in there?

majorz:
> Oh, I was assuming it'd be easy to re-use the regexes we already have in there?

This one would be very similar to the one extracting the `name` field, but that one is the hairiest of it all :)

I am almost done now, and looking at the result I am more happy with the current state, since it reduces the error checking outside of the `replace` function even more.

majorz:
@lurch all of the discussed changes are ready. Please note that I have not touched the python `replace`, since this one should be better handled in a separate issue.

tests/presets.spec.js
```diff
@@ -740,6 +740,445 @@ describe('Presets', function() {
...
+
+            m.chai.expect(fs.readFileSync(this.cargoToml, 'utf8')).to.equal([
+              '[package]',
+              'name = "foo"',
+              'version = "1.2.3"',
+              '',
+              '[dependencies.bar]',
+              'version = "2.0.0"',
+              ''
+            ].join('\n'));
```
lurch: Given that the "before-string" and the "after-string" are identical in this particular test-case, would it be worth storing it in a single const-variable and using that in both places?

tests/presets.spec.js
```diff
@@ -740,6 +740,445 @@ describe('Presets', function() {
...
+            ''
+          ].join('\n'));
+        });
+
+        afterEach(function() {
+          fs.unlinkSync(this.cargoToml);
+          this.cwd.removeCallback();
+        });
+
+        it('should be able to correct the version', function(done) {
```
lurch: minor nit, but my gut-feeling is that this should say "update the version" rather than "correct the version"

lib/presets.js
```diff
@@ -305,22 +338,57 @@ module.exports = {
...
+            done(new Error(`Package name not found in ${cargoToml}`));
+          } else {
+            done(null, matches[2]);
+          }
+        },
+
+        (packageName, done) => {
+          if (fs.existsSync(cargoLock)) {
+            // Update first `version = "..."` occurrence immediately after `name = "${packageName}"`
+            const cargoLockRe = new RegExp(
```
lurch: Just out of curiosity, why do you construct a seaparte RegExp object here, but then just pass in the regex directly on line 386 ?

lib/presets.js
```diff
@@ -27,9 +27,42 @@ const path = require('path');
...
+        done(new Error(`Regex does not match ${file}`));
+      }
+    },
+
+    (contents, done) => {
+      const updated = contents.replace(pattern, replacement);
+      done(null, updated);
+    },
+
+    _.partial(fs.writeFile, file)
```
lurch: Given that we explicitly set a file-encoding for `readFile`, should we also explicitly set a file-encoding for `writeFile` ?

tests/presets.spec.js
lurch: ...and to make this test more comprehensive, should it be updating both Cargo.lock and Cargo.toml that had `version = "not-a-semver"` ?

tests/presets.spec.js
```diff
@@ -740,6 +740,445 @@ describe('Presets', function() {
...
+
+          fs.writeFileSync(this.cargoToml, [
+            '[package]',
+            'name = "foo"',
+            'version = "1.0.0"',
+            '',
+            '[dependencies.bar]',
+            'version = "2.0.0"',
+            ''
+          ].join('\n'));
```
lurch: Along similar lines to what you're testing in Cargo.lock here, there should probably also be a separate test with Cargo.toml set to:
```js
fs.writeFileSync(this.cargoToml, [
  '[package]',
  'name = "foo"',
  '',
  '[dependencies.bar]',
  'version = "2.0.0"',
  ''
].join('\n'));
```
?

tests/presets.spec.js
```diff
@@ -740,6 +740,445 @@ describe('Presets', function() {
...
+        });
+
+        afterEach(function() {
+          fs.unlinkSync(this.cargoToml);
+          this.cwd.removeCallback();
+        });
+
+        it('should be able to update the version (and preserve the single-quotes)', function(done) {
```
lurch: Starting to get a bit nit-picky, but is it worth a test with double-quotes in Cargo.toml but single-quotes in Cargo.lock? \*shrug\*

tests/presets.spec.js
lurch: ...actually, I guess there's nothing that restricts the version we're _setting_ to be a semver either, so perhaps this test (or a new one) should be `it('should work with non-semver versions', ` and then update the version from `not-a-semver` to `also-not-a-semver` ? :smiley:

lib/presets.js
```diff
@@ -27,9 +27,42 @@ const path = require('path');
...
+  async.waterfall([
+    (done) => {
+      fs.readFile(file, 'utf8', done);
+    },
+
+    (contents, done) => {
+      if (pattern.test(contents)) {
+        done(null, contents);
+      } else {
+        done(new Error(`Regex does not match ${file}`));
```
lurch: To match the function parameter names, should this say "Pattern" instead of "Regex"?

majorz: Will modify that.

lib/presets.js
majorz: The encoding defaults to 'utf-8' when writing. It is different than readFile..

lib/presets.js
majorz: This is a leftover from the previous modifications. I will move it inside the function call.

tests/presets.spec.js
majorz: I will extend the test to include Cargo.lock as well.

tests/presets.spec.js
majorz: Right, I have not thought of it. Nice idea.

tests/presets.spec.js
majorz: Let's try that!

tests/presets.spec.js
majorz: We have a similar test, I will move it next to this one and extend it to include the `dependencies.bar` section.

lib/presets.js
majorz: 'utf8', sorry

lib/presets.js
```diff
@@ -305,22 +338,57 @@ module.exports = {
...
+          } else {
+            done(null);
+          }
+        },
+
+        (done) => {
+          // Update first `version = "..."` occurrence immediately after `[package]`
+          replace(
+            cargoToml,
+            /(\[package\][^]+?version\s*=\s*)("|').*?\2/m,
```
lurch: I suspect that when you add the other test we talked about, the `[^]` above will need to be changed to `[^[]`

tests/presets.spec.js
majorz: It does not work with non-sem-versions. We have a `semver.clean(version)` call at the beginning of the function. This means however that we need to have a test for this.

lib/presets.js
majorz: Nice catch!

majorz:
I recommitted again after going though the items. I moved the different tests around so that the regular cases are on top, and then the misconfigurations after them. Also I grouped some of the tests, since they were either operating on the same input, or could be made to do that.

tests/presets.spec.js
lurch: > It does not work with non-sem-versions. We have a semver.clean(version) call at the beginning of the function.

I guess _strictly_ speaking we should only do that if `https://github.com/resin-io/versionist#incrementversion-functionstring` is set to `semver`. But chances are that `versionist` will never get used with non-semver versioning, so we _probably_ don't need to worry about that.

OTOH, do we actually need `semver.clean`? Will versionist ever generate a 'non-clean' semver version?

majorz: That's interesting. I will trace through the code what the situation is tomorrow and get back to you on this. It looks like `incrementVersion` is not respected. I initially copied this from the `npm` preset, so it should have the same issue likewise (+ the new python handling).

majorz: @lurch @hedss I finished looking into the code:

We do not support non-semver based schemes. If we need to support that we have to add more hooks: e.g. one for reading current version, etc.

`incrementVersion` is for augmenting semver versions in a custom way. When the feature was added the description reads: `This hook can be declared to customise how a version is incremented given an increment level. It can be handy in some edge cases, and defaults to incrementing a version based on semver rules.`.

There are not any tests/usages of `incrementVersion` throughout the code which use a custom function, so it looks like it was just a nice-to-have idea. My recommendation is for removing it as a customization option, unless we have projects that are practically using it. I am not sure that should be the case though, as semver itself provides a lot of extra flexibility, e.g. `semver.inc('1.2.0', 'prepatch') == '1.2.1-0'` or `semver.inc('1.2.3-alpha.0', 'prerelease') == '1.2.3-alpha.1'`.

Our update presets receive already cleaned versions after the incrementation step. This is how `semver.inc` works itself: `semver.inc('v5.3.4', 'patch') == '5.3.5'`. My thinking is that in the update functions we should accept only clean versions and treat any unclean version as an error.

Thinking about that further: a clean version has to be produced at the very first step while reading current version from the repo and all consecutive operations should expect a clean version and should treat unclean input as error (in terms of being more defensive towards the input correctness).

Given all of this, I think those problems should be treated in separate issues and we can leave these aspect as-is in this particular commit. Looking forward to hearing your opinions and ideas.

majorz:
I left a comment in one of the review threads, but it is quite buried, so I am placing a link here: https://github.com/resin-io/versionist/pull/104#discussion_r165018621

tests/presets.spec.js
lurch: I'm not sure... The README says "Versionist is non-opinionated. It adapts to your commit practices" so I dunno if we want to explicitly restrict it to only operating on semvers, or if we want to "leave the door open" for other versioning schemes in future? I think the original intention of versionist was for it to be a totally generic tool that any project could use, but I don't know if it's ever been used outside of Resin?
@jviotti As the original author, what do you think about restricting versionist to be semver-only and getting rid of the `incrementVersion` option?
Maybe this is something that should be discussed in a process call @lekkas ?
(For the record, I have no strong opinions either way. Although IMHO it makes sense for the code in the update presets to be as simple as possible, so if there's anything that's being done in _every_ update preset, then perhaps it makes sense to move that code outside the update presets.)

PR #113, opened 2018-02-08
Add a `quoted` updateVersion preset, which uses a regex to update the version in arbitrary (text) file-types

### LINKED ISSUE #110: Could have a `updateQuotedVersion` preset for `updateVersion`
As initially discussed in https://github.com/resin-io/versionist/issues/84#issuecomment-361959222 (so I won't bother copy'n'pasting the info).

This would be a generic 'version-update' mechanism that would work with _any_ type of text file (i.e. it'd be language / framework independent), as long as the version is contained between single-quote `'` or double-quote `"` characters.

And it probably needs a better name than `quotedPatternReplace` :laughing:

lurch:
Change-type: minor
Fixes #110 

I had a few spare hours this evening, so decided to have a go at some `versionist` hacking :grinning: 

IMHO it works pretty nicely...


lib/presets.js
```diff
@@ -428,6 +428,63 @@ module.exports = {
...
+
+      const cleanedVersion = semver.clean(version);
+
+      if (!cleanedVersion) {
+        return callback(new Error(`Invalid version: ${version}`));
+      }
+
+      replace(
+        updateFile,
+        new RegExp('(' + RegExp(options.regex).source + ')("|\').*?\\2', options.regexFlags),
```
majorz: What is the reasoning behind piping `options.regex` through `RegExp(...).source`, instead of using it directly for concatenation?

lib/presets.js
```diff
@@ -428,6 +428,63 @@ module.exports = {
...
+     * @param {String} options.file - file to modify
+     * @param {String} options.regex - regex leading up to the quoted version string
+     * @param {String} [options.regexFlags] - any modifier flags to apply to the regex
+     * @param {String} cwd - current working directory
+     * @param {String} version - version
+     * @param {Function} callback - callback (error)
+     * @returns {null}
+     *
+     * @example
+     * presets.updateVersion.quoted({file: "myfile.h", regex: /^VERSION\s+=\s+/}, process.cwd(), '1.0.0', (error) => {
```
majorz: 1. Using `\s*` instead of `\s+` could be a better choice here?
2. Should `regexFlags: 'm'` be used in conjunction with `^`?
3. Single quotes for `file: "myfile.h"` -> `file: 'myfile.h'`

lib/presets.js
```diff
@@ -428,6 +428,63 @@ module.exports = {
...
+    /**
+     * @summary Update quoted version immediately following a regex
+     * @function
+     * @public
+     *
+     * @param {Object} options - options
+     * @param {String} [options.baseDir] - directory to prefix onto options.file
+     * @param {String} options.file - file to modify
+     * @param {String} options.regex - regex leading up to the quoted version string
+     * @param {String} [options.regexFlags] - any modifier flags to apply to the regex
```
majorz: Something in those lines should point the user where to look for more information on regex flags: `modifier flags as used in RegExp`? It is also clear that way that this is the same argument, not something different.

tests/presets.spec.js
```diff
@@ -1248,6 +1248,809 @@ describe('Presets', function() {
...
+            ].join('\n'));
+
+            done();
+          });
+        });
+
+        it('should be able to update the version using a RegExp object', function(done) {
+          presets.updateVersion.quoted({
+            file: 'File.ver',
+            regex: RegExp('version\\s*=\\s*')
```
majorz: What will happen if we pass `RegExp(/^version\s*=\s*/, 'm')` here - it won't work, right?

I think there could be friction with flags defined in `RegExp` and `regexFlags`. Now that I am thinking maybe we should remove `regexFlags` completely, but provide a default value for flags if a string is passed (e.g. 'm'). Or if `RegExp` is passed, then use its flags. What do you think?

lib/presets.js
```diff
@@ -428,6 +428,63 @@ module.exports = {
...
+     * presets.updateVersion.quoted({file: "myfile.h", regex: /^VERSION\s+=\s+/}, process.cwd(), '1.0.0', (error) => {
+     *   if (error) {
+     *     throw error;
+     *   }
+     * });
+     */
+    quoted: (options, cwd, version, callback) => {
+      _.defaults(options, {
+        baseDir: '.',
+        regexFlags: ''
```
majorz: I am not sure this is really needed as `undefined` is the default for `RegExp`. But then having a value can be a better practice for default options.

lib/presets.js
```diff
@@ -428,6 +428,63 @@ module.exports = {
...
       });
+    },
+
+    /**
+     * @summary Update quoted version immediately following a regex
+     * @function
+     * @public
+     *
+     * @param {Object} options - options
+     * @param {String} [options.baseDir] - directory to prefix onto options.file
```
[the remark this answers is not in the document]
lurch: > Since we are going to use `baseDir` as a generic parameter for all presets, I think we need a more generic description than directory to prefix onto options.file.

Good point - not all presets will have a `file` option. What about "relative directory to append to cwd" ?

README.md
```diff
@@ -617,6 +617,17 @@ This preset updates the `version` property of `$CWD/Cargo.toml` (and also `$CWD/
...
+
+This preset updates the (double-quoted or single-quoted) version string
+immediately following `regex` in the given `file`. For example a file
+containing `project.version = "1.4.2";` could be updated by using the
+`quoted` prefix with a `regex` of `/project\.version\s*=\s*/`. You can
+optionally provide `regexFlags` to modify the `regex`'s behaviour (if using
+the `^` start-of-line character in the `regex`, don't forget to add `m` to
+`regexFlags`). You can optionally provide a `baseDir` to modify where it looks
+for `file` (it will update `$CWD/$baseDir/$file`).
+
```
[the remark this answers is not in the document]
lurch: I dunno - given how short all the other preset explanations are (and IIRC nobody has yet complained about the lack of documentation), I wasn't sure how much depth to go into in the README... :-/

It's not really an excuse, but personally when I can't figure out how an option works, I either look for an existing example elsewhere to copy from, or I have a look at what the unit-tests are doing ;-)

lib/presets.js
```diff
@@ -428,6 +428,63 @@ module.exports = {
...
+    quoted: (options, cwd, version, callback) => {
+      _.defaults(options, {
+        baseDir: '.',
+        regexFlags: ''
+      });
+
+      if (path.isAbsolute(options.baseDir)) {
+        return callback(new Error('baseDir option can\'t be an absolute path'));
+      }
+      if (!options.file) {
```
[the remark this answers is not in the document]
lurch: Oh probably - I'm still quite a JS newbie myself. I'll make that change.

lib/presets.js
lurch: If you concatenate it directly, it works for strings but not for regex literals
```js
$ node
> '(' + 'version\\s*=\\s*' + ')("|\').*?\\2'
'(version\\s*=\\s*)("|\').*?\\2'
> '(' + /version\s*=\s*/ + ')("|\').*?\\2'
'(/version\\s*=\\s*/)("|\').*?\\2'
> 
> '(' + RegExp('version\\s*=\\s*').source + ')("|\').*?\\2'
'(version\\s*=\\s*)("|\').*?\\2'
> '(' + RegExp(/version\s*=\s*/).source + ')("|\').*?\\2'
'(version\\s*=\\s*)("|\').*?\\2'
```
See also https://stackoverflow.com/a/185529

(although this is the first time I've used regexes like this in javascript, so there might well be a better way of doing this!)

lib/presets.js
lurch: All very good points :+1:

README.md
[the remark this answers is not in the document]
majorz: I think this could be handled later in a separate issue.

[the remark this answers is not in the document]
majorz: I think this could be handled later in a separate issue.

lib/presets.js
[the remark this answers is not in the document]
majorz: Yes. Later on when `baseDir` is added globally its purpose could be clarified in the README.

tests/presets.spec.js
lurch: The 'm' specified inside the `RegExp` constructor would be lost, as https://stackoverflow.com/a/185529 notes.

> Or if RegExp is passed, then use its flags.

Good idea - I now merge the flags from `regex` and `regexFlags` :smiley:

lib/presets.js
lurch: Nope, default is an empty string
```js
$ node
> RegExp().flags
''
```

(and if I change this to `undefined` my new regex-flags-merging code breaks :laughing: )

lib/presets.js
lurch: ...except if I change it to `/^VERSION\s*=\s*/` then the interpreter sees the `*/` as the end of the comment-block, so I've switched it back to `/^VERSION\s+=\s+/` :wink:

lurch:
Just pushed new updates

majorz:
Huh, I do not see a button on how to approve from my phone but will click it when I get back home..

tests/presets.spec.js
majorz: Merging is a nice option.

tests/presets.spec.js
```diff
@@ -1248,6 +1248,891 @@ describe('Presets', function() {
...
+              ''
+            ].join('\n'));
+
+            done();
+          });
+        });
+
+        it('should be able to update the version when adding the subdirectory to the filename', function(done) {
+          presets.updateVersion.quoted({
+            file: 'subdirectory/File.ver',
```
lurch: I guess this should have been `path,join('subdirectory', 'File.ver')` - oh well, doesn't matter :)

majorz:
@lurch it looks like the label is not applied for some reason.

majorz:
Ah, and what is the policy for squashing the commits? Will versionbot do that?

lurch:
The label _was_ still applied, just looks like VersionBot had fallen asleep or something? :laughing:

hedss:
@lurch Can you rebase into a single commit, as per-PR guidelines, please, before merging?

lurch:
> Can you rebase into a single commit

Done.

> as per-PR guidelines, please

Sorry, I hope I've not messed anything up. I didn't realise the VersionBot wouldn't auto-squash if needed.
 (Off-topic, is not allowing non-squashed PRs something that VersionBot could also check for / enforce?)

I think Etcher uses a different set of "merge guidelines"? https://www.flowdock.com/app/rulemotion/r-etcherprv/threads/OH5oivc8tStaTKsN76oVwGD7zMi

hedss:
@lurch Yeah, `etcher` and `resin-os` are the two at the moment with different guidelines. When VB2 comes along I suspect everything will be unionified.

lurch:
Hurrah for onions :wink:

lurch:
Yippee! Couple of mis-steps but I got there in the end :tada:

PR #115, opened 2018-03-27
Adds support for overidding `getIncrementLevelFromCommit` with command line flags `--patch|-p` `--minor|-m` `--major|-M`

danielmahon:


danielmahon:
this makes it easier to create npm scripts to handle releases, If I'm not mistaken, currently I would need 3 different config files or pass the version in the commit message (which I don't need to do). With this update I can create scripts like the following:

package.json
```json
...
  "scripts": {
    "release:minor": "versionist --minor",
    "release:patch": "versionist --patch",
    "release:major": "versionist --major"
  },
...
```
Im also using the `versionist-plugins` to handle commiting, tagging, and pushing, but If I wasn't I could chain commands here

danielmahon:
here is the corresponding config:

**versionist.conf.js**
```javascript
const plugins = require('versionist-plugins');

module.exports = {
  updateVersion: [plugins.git.commit, plugins.git.tag, plugins.git.push],
  template: [
    '## v{{version}} - {{moment date "Y-MM-DD"}}',
    '',
    '{{#each commits}}',
    '- {{capitalize this.subject}}',
    '{{/each}}',
  ].join('\n'),
  getGitReferenceFromVersion: version => `v${version}`,
};
```

danielmahon:
Thoughts on implementing this? Do I need to fix something?

brownjohnf:
@resin-ci retest

PR #117, opened 2018-06-19
Add get-current option

nazrhom:
Returns what is considered the latest documented version.

change-type: minor


bin/cli.js
```diff
@@ -118,8 +128,13 @@ async.waterfall([
 
       return documentedVersions;
     });
+    const latest = semver.getGreaterVersion(versions);
+    if (argv['get-current']) {
+      console.log(latest);
+      return callback(STOP);
```
hedss: Don't understand the use of this `Stop` error class, why not just `callback(null)`?

bin/cli.js
```diff
@@ -65,10 +71,14 @@ const argv = yargs
   })
   .options({
     current: {
```
hedss: Ideally this needs now to become `set-current`. It becomes a major change, but otherwise the naming convention isn't very good.

bin/cli.js
```diff
@@ -65,10 +71,14 @@ const argv = yargs
...
-      describe: 'current version',
+      describe: 'set current version',
       string: true,
       alias: 'u'
     },
+    'get-current': {
+      describe: 'get current version',
```
hedss: ` and exit without making changes` is probably worth appending here.

bin/cli.js
nazrhom: This is to avoid printing`Done` at the end, since we want to use this feature in automation it doesn't make sense to add extra input we have to throw away.

bin/cli.js
nazrhom: I agree, alternatives could be `--latest-documented-version` to avoid confusion with `current`. Another idea could be to add a `get` command that could accept `get current` and `get reference` which would return respectively the latest documented version, and latest git reference

lurch: Wouldn't "get reference" be a job for the `git` command, rather than the `versionist` command?

nazrhom: In this case the reference you get would be the result of `getChangelogDocumentedVersions`, followed by `getGitReferenceFromVersion`, its not actually walking your git history or anything like that, in fact it could return the name of a reference that does not exist yet.

lurch: Ahhh, I'd forgotten about https://github.com/resin-io/versionist/blob/master/README.md#getgitreferencefromversion

bin/cli.js
hedss: We do, but it could also be used as a CLI command still. So I'd either propose removing `Done` altogether or not worrying about it! If an error is percolated up and you're running this as an `exec()` process, for example, you'll still get an error code returned rather than having to worry about `stdout` which you can just ignore!

bin/cli.js
hedss: Yeah, then let's have a `get <subcommand>`. :+1:

bin/cli.js
lurch: Or perhaps another command-flag to turn off the "Done" output? :man_shrugging:

bin/cli.js
lurch: :wave:

nazrhom:
@hedss actually the reason why we need the `Stop` class is to exit early from `async.waterfall`. If we call `callback(null)` then the next action in the waterfall will be called. This way we can distinguish if we are cutting the waterfall early because of an actual error, or because that is the desired behaviour.

bin/cli.js
```diff
@@ -93,6 +99,14 @@ const argv = yargs
       alias: 'v'
     }
   })
+  .command('get <target>', 'get latest documented version or reference', (yargsGet) => {
+    yargsGet.positional('target', {
+      alias: 't',
```
lurch: Does this mean that I could also do `versionist -t reference` ?

lurch:
It's just occurred to me that this PR should probably update the README as well? :slightly_smiling_face:

bin/cli.js
nazrhom: Good catch, options declared inside a command apply only for the command, so anything in there is only valid when you do `versionist get ...`. Having said this, I could not find way to use the alias in a sub-command so I removed it and added `choices` instead

README.md
```diff
@@ -91,6 +94,12 @@ You can use this option to pass a custom location to `versionist.conf.js`.
 You can use this option to perform a dry run to see what changes _would_ be
 made by versionist, without actually changing any files.
 
+### `get <target>`
+
+You can use this option to return the latest documented version (according to `getChangelogDocumentedVersions`) or a git reference which is produced by applying `getGitReferenceFromVersion` to the latest documented version.
+This command will not run versionist.
```
lurch: LOL, "This command will not run versionist" sounds a little funny :laughing: Did you mean something like "This option will not cause any changes" ?

tests/cli/utils.js
```diff
@@ -48,7 +48,11 @@ exports.callVersionist = (opts) => {
   const cliPath = path.join(__dirname, '..', '..', 'bin', 'cli.js');
 
   const configString = _.reduce(opts, function(result, value, key) {
-    result += ` --${key}=${value}`;
+    if (key === 'cmd' || key === 'command') {
+      result = ` ${value} ${result}`;
```
lurch: Would this line be simpler as
```
result += ` ${value}`;
```
?
(also makes it more similar to the 'else' part).

nazrhom: The order would be flipped in that case, we want the command to appear first and the options after that.

README.md
nazrhom: Hehehe that is true, I will update it :)

tests/cli/utils.js
lurch: Oh, does it make a difference? (I'm not that familiar with yargs, but there's nothing in your updated README that says "commands" need to be specified before "options" ?)

README.md
```diff
@@ -91,6 +94,12 @@ You can use this option to pass a custom location to `versionist.conf.js`.
 You can use this option to perform a dry run to see what changes _would_ be
 made by versionist, without actually changing any files.
 
+### `get <target>`
+
+You can use this option to return the latest documented version (according to `getChangelogDocumentedVersions`) or a git reference which is produced by applying `getGitReferenceFromVersion` to the latest documented version.
```
lurch: Should this say "this command" instead of "this option" ?

tests/cli/utils.js
nazrhom: I tested some options and it seems to not matter. I am not sure if this will always hold in the future though (if we extend versionist in the future with more complex commands which take extra options).
I clarified in the readme which options are accepted for the `get`command

lurch: Just to clarify - I was wondering if it might be worth updating the Usage string in the README to something like:
```
Usage: versionist [COMMAND] [OPTIONS]
```
? :man_shrugging:

hedss:
@nazrhom I still don't like the use of a new class to signal the error. Why not just throw an error with a code and check against it. Using a class instance to do this just seems wrong to me. Apart from that, all looks fine. :)

bin/cli.js
```diff
@@ -118,8 +130,17 @@ async.waterfall([
...
+    if (argv.target === 'version') {
+      console.log(latest);
+      return callback(STOP);
+    }
 
-    const gitReference = argv.configuration.getGitReferenceFromVersion(semver.getGreaterVersion(versions));
+    const gitReference = argv.configuration.getGitReferenceFromVersion(latest);
+    if (argv.target === 'reference') {
+      console.log(gitReference);
+      return callback(STOP);
```
hedss: Sorry, I'm probably not being very clear here. Why not define the string code (if we're not using a number) at the top of the file:
```
const stopError = 'E_STOP`;
```
Then in the code just do:
```
return callback(throw new Error(stopError));
```
and then when testing, just:
```
if (!(error.message === stopError))
```
?

It cuts down the amount of code and doesn't leave you creating an object immediately which might never get used.

nazrhom: this is how I had it initially, but I was not that happy with checking the message to determine the error type... `error.code` seems better for this case. I initialised the error on the top level to avoid having to set the code in every occasion you want to use stop. Maybe a compromise could be to use `.code` but only set it after constructing the error in the if blocks?

hedss: I'm not sure I understand. `.code` is not a default property of a vanilla JS `Error`, whereas `.message` is, which doesn't require any extra code to set it as it's part of the constructor. If we were using a subclass inheriting from this (such as `typed-error`) because we needed to (like we do in a lot of backend components), I'd agree, but for a one-off to break the waterfall I'm not sure I see any benefit to it.

However, if you want to keep it like this, I'll approve.

Issue #83, opened 2017-05-15
Requires an API interface to drive directly as a module


hedss:
VersionBot calls `versionist` as an executable. This is not ideal, as there are instances where we would like to pass different options around.

dt-rush:
Hey folks,

I took some time out yesterday and this morning to work on this. It was quite a bit of a task to cut it free, but I've done it. It's now callable via a All tests are passing and it just needs a look-over.

PR #118, opened 2018-08-12
separating out a programmatic interface from the CLI code

dt-rush:
Refactored the running of the tool into a function of the `versionist` module, added support for it to be called via an opts object. Also some minor improvements like adding a test for `showErrorAndQuit` and factoring out helper functions from the cli script into `versionist`.

Change-Type: minor
Connects-to: #83
---- Autogenerated Waffleboard Connection: Connects to #83


brownjohnf:
@resin-ci retest

dt-rush:
I noticed there had been new commits on master since I sent this PR. I'll merge them into this branch and then we can merge it into master.

dt-rush:
@nazrhom  @hedss @jviotti @imrehg 

I've just now merged the changes that happened on master between when this PR was made. The other two PR's currently waiting can also be merged quite simply by considering the following:

If merge conflicts arise for `bin/cli.js`, this is because this branch moved most of the business logic in `bin/cli.js` into `lib/versionist.js` (`bin/cli.js` becoming more properly simply concerned with cli logic like arg parsing and output), renaming `argv` to `parsedArgs` (since `argv` corresponds to the unparsed argv object, properly speaking, while the `argv` of the `cli.js` logic represents the parsed args, and it's more clear to name it as such).

nazrhom:
Hey @dt-rush thanks for the contribution and sorry for the delayed response.

I like this, personally I am only unsure about exposing the API with a promise interface. The reason being that most of the other internals we export are callback based; users can always promisify the whole thing, should they want to.

tests/cli/cases/dry-mode.js
```diff
@@ -27,6 +27,8 @@ shelljs.cd(TEST_DIRECTORY);
 
 const originalChangelog = shelljs.cat('CHANGELOG.md').stdout;
 
+console.log(`originalChangelog: ${originalChangelog}`);
+
```
nazrhom: NIT: Is this supposed to be here?

dt-rush: Great catch. I'll remove that log line and produce a callback version which is more idiomatic.

dt-rush: Pushed just now.

dt-rush:
@nazrhom @imrehg this branch is now up to date with master, having merged the changes that happened since this branch was first created. It passes all tests and I've been meticulous about porting the code in the commits to the appropriate places in the refactor. I'm not opposed to continuing to do this... but we should merge sooner rather than later, to avoid this getting stuck in this state forever.

PR #120, opened 2018-08-23
lib/semver: Modify getGreaterVersion

nazrhom:
Extend getGreaterVersion to work for extended-semver strings.

Change-type: minor
Signed-off-by: Giovanni Garufi <giovanni@resin.io>


lib/semver.js
```diff
@@ -161,6 +161,35 @@ exports.checkValid = (version) => {
...
+ * a = b => 0
+ * a < b => -1
+ *
+ * @example
+ * compareExtendedSemver('2.1.1', '2.1.0')
+ * > 1
+ * compareExtendedSemver('2.1.1', '2.1.0+rev1')
+ * > 1
+ * compareExtendedSemver('2.1.1+rev1', '2.1.1+rev2')
+ * > -1
```
lurch: How does it behave when doing e.g. `compareExtendedSemver('2.1.1', '2.1.1+rev1')`

nazrhom: This would still be -1

nazrhom: Added a test case for that

PR #122, opened 2018-09-13
presets: Change presets to match default config

nazrhom:
Change-type: major
Signed-off-by: Giovanni Garufi <giovanni@resin.io>


lib/presets.js
```diff
@@ -42,6 +43,16 @@ const authenticate = () => {
   });
 };
 
+const isIncrementalCommit = (changeType) => {
+  return Boolean(changeType) && changeType.trim().toLowerCase() !== 'none';
```
hedss: Really should be parentheses around the second test.

lurch:
Does changing the presets to match the default config mean that Versionist is now slightly _less_ non-opinionated ? :wink:

nazrhom:
@hedss yup. Btw I had to revert the parenthesis in 
```
  return Boolean(changeType) && changeType.trim().toLowerCase() !== 'none';
```
because eslint was complaining

nazrhom:
@resin-ci retest

nazrhom:
@resin-ci retest

PR #126, opened 2018-11-15
Fix nested-changelog range

nazrhom:
Change-type: patch


lurch:
Seems this change was already included in #127 ?

Issue #103, opened 2018-01-22
The `npm` updateVersion preset should also update npm-shrinkwrap.json and package-lock.json


lurch:
If a NodeJS project is using versionist, then it can use the `npm` [updateVersion preset](https://github.com/resin-io/versionist#updateversion) to automatically update the version number in `package.json`.
However _if_ the project also has `npm-shrinkwrap.json` or `package-lock.json` files, versionist should also update the version number stored in these file(s).

lurch:
Implemented in #131

PR #131, opened 2019-02-26
Check package-lock.json and npm-shrinkwrap.json

thundron:
Change-type: patch
Signed-off-by: Lorenzo Alberto Maria Ambrosi <lorenzothunder.ambrosi@gmail.com>

### DIFF
--- a/lib/presets.js
+++ b/lib/presets.js
@@ -588,20 +588,46 @@ module.exports = {
     npm: (options, cwd, version, callback) => {
       const cleanFn = getCleanFunction(options);
       const packageJSON = path.join(cwd, 'package.json');
+      const packageLockJSON = path.join(cwd, 'package-lock.json');
+      const npmShrinkwrapJSON = path.join(cwd, 'npm-shrinkwrap.json');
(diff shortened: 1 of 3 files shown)


lurch:
Looks like this fixes #103 ? :smiley:

nazrhom:
@thundron the changes looks good. I would only add a test case for it in https://github.com/balena-io/versionist/blob/master/tests/presets.spec.js#L710

lurch:
Would a project ever have _both_ a `npm-shrinkwrap.json` and a `package-lock.json` (if so, should there be a test for that too?), or are they effectively mutually exclusive?

thundron:
@lurch There's no reason to have both afaik, they serve different purposes wrt dependencies:
`npm-shrinkwrap.json` is used to respect deep sub-dependencies too and is published to `npm` as well, whereas `package-lock.json` is not and just respects top-level dependencies.
The shrinkwrap file is also compatible with older versions of npm
You can find more [here](https://stackoverflow.com/questions/44258235/what-is-the-difference-between-npm-shrinkwrap-json-and-package-lock-json)

PR #136, opened 2019-04-17
do not require "Update" to be title-case in commit body

wrboyce:
Change-type: patch
Signed-off-by: Will Boyce \<will@balena.io\>


nazrhom:
@wrboyce we need to update https://github.com/balena-io/github-events-resource/blob/master/Dockerfile#L14 after the new version is released. After the update is merged the changes will be available in the CI

wrboyce:
Thanks @nazrhom!

wrboyce:
@nazrhom is it worth adding a changelog.yml to this repo so `github-events-resource` gets a more informative changelog?

PR #139, opened 2019-04-17
Multiple footers with the same key will be parsed

nazrhom:
with a number appended to them

Change-type: minor

### DIFF
--- a/tests/tags.spec.js
+++ b/tests/tags.spec.js
@@ -238,6 +238,22 @@ describe('Tags', function() {
         bar: 'baz'
       });
     });
+
+    it('should not overwrite footers with the same key', function() {
+      const footer = tags.parseFooterTagLines([
+        'Foo: bar',
+        'Foo: bar',
+        'Bar: baz',
+        'Foo: baz'
+      ]);
+
+      m.chai.expect(footer).to.deep.equal({
+        Foo: 'bar',
+        Foo1: 'bar',
+        Bar: 'baz',
+        Foo2: 'baz'
+      });
+    });
   });
 
 });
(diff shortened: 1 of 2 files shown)


tests/tags.spec.js
```diff
@@ -238,6 +238,22 @@ describe('Tags', function() {
...
+
+    it('should not overwrite footers with the same key', function() {
+      const footer = tags.parseFooterTagLines([
+        'Foo: bar',
+        'Foo: bar',
+        'Bar: baz',
+        'Foo: baz'
+      ]);
+
+      m.chai.expect(footer).to.deep.equal({
```
wrboyce: Would `{Foo: ['bar', 'bar', 'baz'], Bar: 'baz'}` not be a better output?

tests/tags.spec.js
```diff
@@ -238,6 +238,22 @@ describe('Tags', function() {
         bar: 'baz'
       });
     });
+
+    it('should not overwrite footers with the same key', function() {
+      const footer = tags.parseFooterTagLines([
+        'Foo: bar',
+        'Foo: bar',
+        'Bar: baz',
+        'Foo: baz'
```
lurch: Might also be worth having a test with 
```
[
    'Foo: bar',
    'Foo1: moo',
    'Foo: bar',
    'Bar: baz',
    'Foo: baz'
]
```
to check that the re-numbering doesn't overwrite existing tags?

tests/tags.spec.js
lurch: I guess that wouldn't allow you to tell that `'Bar: baz'` occurs before the last `'Foo: baz'`, but I dunno if that's important or not?

I guess a downside to @nazrhom 's proposed output is that you won't be able to tell the difference between inputs of
```
[
    'Foo: bar',
    'Foo: bar',
    'Bar: baz',
    'Foo: baz'
]
```
and
```
[
    'Foo: bar',
    'Foo1: bar',
    'Bar: baz',
    'Foo2: baz'
]
```
but again I've got no idea if that's important or not.

PR #140, opened 2019-04-29
Use resin semver

nazrhom:


tests/semver.spec.js
```diff
@@ -313,6 +313,30 @@ describe('Semver', function() {
...
+      const greater = semver.getGreaterVersion([
+        'balenaOS 2019.04',
+        'balenaOS 2019.04.1',
+        'balenaOS 2019.04.2',
+        'balenaOS 2019.04.3',
+        'balenaOS 2019.4.2',
+        'balenaOS 2019.4.1'
+      ]);
+
+      m.chai.expect(greater).to.equal('balenaOS 2019.04.3');
```
lurch: Just out of curiosity, how would 'balenaOS 2019.04.3' be ordered against 'balenaOS 2019.4.3' ?

nazrhom: @lurch they would be considered equal, cfr. https://github.com/balena-io-modules/resin-semver/blob/master/test/index.spec.ts#L19

PR #142, opened 2019-05-02
Typescript conversions

Page-:


src/markdown.ts
```diff
@@ -14,14 +14,15 @@
...
-
 /**
  * @module Versionist.Markdown
  */
 
-const _ = require('lodash');
-const markdown = require('markdown').markdown;
+import * as _ from 'lodash';
+// tslint:disable-next-line:no-var-requires
+const { markdown } = require('markdown') as {
```
Page-: This module hasn't been updated in 6 years and is unmainted (https://github.com/evilstreak/markdown-js#notice-unmaintained) so I'm not sure why we use it specifically but I just hacked together these very basic types for now rather than replacing it

Issue #107, opened 2018-01-29
versionist should support updating several files at once


lurch:
Spin-off from a conversation in https://github.com/resin-io/reconfix/pull/40

As seen in the linked reconfix PR, there might be circumstances in which we want to update several different files at once, bumping them all to the same new version. My suggestion for doing this is to modify the `updateVersion` parameter of `versionist.conf.js` so that instead of accepting a function or a preset-string or a preset-object, it instead accepts an array of functions and/or preset-strings and/or preset-objects. So that if you specified something like the below in `versionist.conf.js` (and assuming that #105 and #103 also get implemented)...
```js
updateVersion: [
    'cargo',
    {
        preset: 'cargo',
        targetDir: 'bindings/node/native'
    },
   'npm'
]
```
...then versionist would bump the versions in each of: `Cargo.toml`, `bindings/node/native/Cargo.toml`, `package.json` and `package-lock.json`.

(Although having looked at this in a bit more detail, I'm unsure if this idea would be complementary-to or conflicting-with #78 )

lurch:
Fixed by #143

Issue #73, opened 2017-02-21
Versionist should continue even if a `package.json` for the `updateVersion` default doesn't exist


hedss:
Currently `versionist` will error if it uses the default `npm` use of `updateVersion` and no `package.json` exists in the repository it's executing in.

What should happen is a warning is returned, but the `CHANGELOG` is still correctly updated.

Without this functionality, every non-Node based repo currently needs a custom `versionist.conf.js` which is far from ideal (see https://github.com/resin-io/resin-containers for an example, which has an empty callback).

lurch:
Is the `updateVersion` setting in resin-containers's `versionist.conf.js` the only difference from `versionist`'s built-in config file? Would it make sense for `versionist` to apply values from its built-in config file for any settings not found in the current directory's `versionist.conf.js`? Or is that a nonsensical suggestion which would be too backwards-incompatible?

hedss:
I don't think this would be a good idea, mainly as it might apply settings for things specifically not required. Our real issue here is that the default assumes NPM and errors on non-NPM. I would expect it to simply not care if there was no `package.json`.

craig-mulligan:
Agreed, yea should just log `no package.json` rather than throwing.

lurch:
So just to clarify, if the (built-in) default `updateVersion` setting for `versionist` is `'npm'`, and there's no `versionist.conf.js` in the current directory, and also no `package.json` in the current directory, then `versionist` should treat `updateVersion` as `(cwd, version, cb) => { cb(); }` ?
Having just had a quick look at the versionist README, rather than overriding `updateVersion` to `(cwd, version, cb) => { cb(); }`, would it be cleaner to instead override `editVersion` to `false`?

And IMHO this behaviour should only be triggered **if** there's no local `versionist.conf.js` and also no local `package.json`. If there is a local `versionist.conf.js` but no local `package.json` then that should error, because there's no reason not to just update the local `versionist.conf.js` to suit.

hedss:
That's a good point; `editVersion` as `false` would have been a lot cleaner!
I also agree that if there is an overriding `versionist.conf.js` which specifies project update and there's no relevant file (such as `package.json`) then that is definitely an error.

So, in conclusion:
* The default internal `versionist.conf.js` should be set to `updateVersion: npm` as it is at the moment. Should no `package.json` be found in the repo using this default, then a verbose warning occurs, but `versionist` completes successfully
[...]

PR #143, opened 2019-05-13
Add mixed preset for updateVersion

nazrhom:
This will attempt an npm/cargo/python/generic update in parallel
ignoring any errors.

Change-type: patch

### DIFF
--- a/src/presets.js
+++ b/src/presets.js
@@ -823,6 +823,7 @@ module.exports = {
      * @summary Update version file
      * @function
      * @public
+     * @returns {null}
      *
      * @param {Object} options - options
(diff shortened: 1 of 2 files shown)


lurch:
Looks like this fixes #107 ?

lurch:
Does it also fix #73 ?

nazrhom:
@lurch it definitely addresses #107. As far as #73 goes, a custom versionist.conf.js is injected by balena-versionist so the motivating issue is already partially resolved. Rather than having updateVersion ignore errors by default I preferred having users opt into this behaviour explicitly via `mixed`.

PR #148, opened 2019-06-24
add script for generating structured changelogs from git history

wrboyce:
Change-type: minor
Signed-off-by: Will Boyce \<will@balena.io\>

### DIFF
--- a/scripts/generate-changelog.sh
+++ b/scripts/generate-changelog.sh
@@ -1,0 +1,109 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2094
+set -eu
+
+if [ "${BASH_VERSINFO[0]}" -lt 4 ]; then
+  echo 'ERROR: bash version >= 4.0 is required' >&2
+  exit 1
+fi
+
+for dep in jq yq; do
+  if ! command -v "${dep}" &>/dev/null; then
+    # shellcheck disable=SC2016
+    echo 'ERROR: `'"${dep}"'` command is not available' >&2
+    exit 1
+  fi
+done
+
+yq_min_version="2.4.0"
+yq_version="$(yq --version | awk '{print $3}')"
+if [ "${yq_version}" != "${yq_min_version}" ] && [ "${yq_version}" = "$(echo -e "${yq_version}\n${yq_min_version}" | sort -V | head -1)" ]; then
+  echo "ERROR: yq version >= ${yq_min_version} is required" >&2
+  exit 1
+fi
+
+REPO_PATH="${1-}"
+if [ -z "${REPO_PATH}" ]; then
+    echo "USAGE: $0 <path>" >&2
+    exit 1
+fi
+
+GH_TOKEN="${GH_TOKEN-}"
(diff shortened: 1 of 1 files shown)


scripts/generate-changelog.sh
```diff
@@ -0,0 +1,89 @@
...
+            while read -r trailer; do
+                key=$(cut -f1 -d':' <<<"${trailer}" | tr '[:upper:]' '[:lower:]')
+                val=$(cut -f2- -d':' <<<"${trailer}" | sed -e 's/^[[:space:]]*//')
+                yq write -i "${tmp}/${version}.yml" '[0].commits['"${i}"'].footers.'"${key}" "${val}"
+                unset 'msg['$((${#msg[@]} - 1))']'
+            done < <(jq -cr '.[] | select(.sha == "'"${sha}"'") | .commit.message' <"${tmp}/pr${pullId}.json" | git interpret-trailers --parse)
+            unset 'msg[0]'
+            yq write -i "${tmp}/${version}.yml" '[0].commits['"${i}"'].subject' "${subject}"
+            body=$(awk '{gsub(/^ +| +$/,"")} NF {print $0}' < <(printf "%s\n" "${msg[@]}"))
+            yq write -i "${tmp}/${version}.yml" '[0].commits['"${i}"'].body' "${body}"
```
imrehg:
```suggestion
            yq write -i "${tmp}/${version}.yml" '[0].commits['"${i}"'].body' -- "${body}"
```
to be able to handle some of the cases when the `body` has text, that `yq` might interpret as flags. The change is from @Page-

imrehg:
@wrboyce I also wonder, this relies on sme external tools such as `yq`, is there a way to specify that somewhere?

roman-mazur:
Guys, I noticed we have cases when `change-type` and `signed-off-by` properties are duplicated because of case difference.
https://github.com/balena-io/balena-supervisor/blob/master/.versionbot/CHANGELOG.yml#L8

Any idea how this can be fixed?

wrboyce:
@roman-mazur This is done by versionist (as opposed to the retrospective generation script) and IIRC it was by design, I'll try and find the Flowdock thread and ping you on it.

nazrhom:
Thanks @wrboyce this looks great. A couple of comments after trying to run this on OSX: 
I noticed the file is not executable.
The first issue shows up at `line 34: declare: -A: invalid option` This is because OSX ships with bash < 4 by default and does not have that option yet. 
Also mapfile is not present by default, maybe we can add a check at the top to see if it is installed?

It doesn't HAVE to be compatible but it would be nice if we make it work cross-platform

wrboyce:
@nazrhom added some checks for dependencies, you can install a compatible version of bash from homebrew (bash went GPLv3 from v4 so Apple won't ship it, personally I welcome zsh becoming the default as of macOS 10.15!).

wrboyce:
cc @imrehg added checks for dependencies and `yq` version.

jviotti:
Worked like charm on Landr

PR #153, opened 2019-09-03
Create parsable changelog if .versionbot folder is present

nazrhom:
Also do not throw an error if CHANGELOG.yml file is empty but just treat
it as an empty array

Fixes: balena-io/balena-versionist#28
Change-type: patch

### DIFF
--- a/tests/cli/cases/empty-parsable-changelog.js
+++ b/tests/cli/cases/empty-parsable-changelog.js
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2019 Balena.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


tests/cli/cases/empty-parsable-changelog.js
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright 2019 Resin.io
```
imrehg: Should this copyright notice be Balena? Since Resin doesn't exists since 2018

imrehg:
A very useful change, should make it easier to set up a new repo with the required files. Have not tested obviously, so this review is with a pinch of salt.

tests/cli/cases/empty-parsable-changelog.js
nazrhom: Updated :P

PR #159, opened 2019-10-09
Create VERSION file if not preset

nazrhom:
Change-type: patch


tests/cli/cases/generic-no-version.js
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright 2016 Resin.io
```
lurch: 2016? Resin.io?? :rofl:

Issue #152, opened 2019-08-20
Dry-run edits machine-generated changelog


xginn8:
If I run versionist in dry-run mode, there are still changes made to the yml changelog:

```
~/balena/device-diagnostics checks-data-split
❯ git status               
On branch checks-data-split
nothing to commit, working tree clean

~/balena/device-diagnostics checks-data-split
❯ versionist -u 4.0.0 --dry
[before-after-hook]: "Hook()" repurposing warning, use "Hook.Collection()". Read more: https://git.io/upgrade-before-after-hook-to-1.4
## v5.0.0 - 2019-08-20

 - Diagnostics: split check and data collection tasks [Matthew McGinn]

Done

~/balena/device-diagnostics checks-data-split*
❯ git status               
On branch checks-data-split
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   .versionbot/CHANGELOG.yml

no changes added to commit (use "git add" and/or "git commit -a")
```

PR #166, opened 2019-12-26
Move editing of CHANGELOG.yml after dry checks

nazrhom:
Change-type: patch
Fixes: #152


lurch:
Maybe it's worth adding a unit-test to ensure that running in dry-run mode doesn't cause any files to get modified? :man_shrugging:

PR #172, opened 2020-06-25
Update dependencies, fixing some warnings

Page-:
Change-type: patch


.gitignore
```diff
@@ -37,6 +37,6 @@ bower_components
 *.pem
 
 # Temporal CLI test data
-tests/cli/temporal
+temporal-tests
```
Page-, line 40: I moved the temporal tests location to make it easier to exclude them from linting

PR #177, opened 2020-07-09
Improve rendering of nested changelogs

Page-:
This causes the generated changelog to be rendered with consistent
blockquote nesting, making it much easier to see the nesting level for
each entry

Change-type: patch


src/presets.js
```diff
@@ -85,7 +86,8 @@ const templateDefaults = [
 
 	'{{#*inline "block-newline"}}',
 		'{{#isnt block ""}}{{block}} {{else}}{{/isnt}}',
-	'{{/inline}}'
+	'{{/inline}}',
+	''
```
Page-, line 90: This was added because it removes a bunch of new lines that I don't want added. I'm not sure why it works just that it does precisely what I need and doesn't seem like it should be harmful?

src/presets.js
```diff
@@ -44,19 +44,20 @@ const templateDefaults = [
 	'{{> block-newline}}',
 	'{{#each commits}}',
 		'{{> render-with-nesting nesting=../nesting block=../block ~}}',
-	'{{/each~}}',
+	'{{/each}}',
+	'{{> block-newline}}',
```
Page-, line 48: This just adds more separation between commits so you get
```
>> ### module-13.0.2
>> #### (2020-06-29)
>>
>> * Commits
>>
>> ### module-13.0.1
>> #### (2020-06-22)
```
instead of
```
>> ### module-13.0.2
>> #### (2020-06-29)
>>
>> * Commits
>> ### module-13.0.1
>> #### (2020-06-22)
```
The actual rendering is identical but the extra spacing helps a lot when looking at the markdown directly

PR #178, opened 2020-07-09
Guarantee the order of nested changelogs

Page-:
Previously the order of nested changelogs was dependent on a race
condition, now it will always be consistent regardless of how long
fetching the nested entries takes

Change-type: patch

### DIFF
--- a/src/presets.js
+++ b/src/presets.js
@@ -152,7 +152,7 @@ const getNestedChangeLog = (
 };
 
 const attachNestedChangelog = (upstreams, commit, callback) => {
-	async.each(
+	async.map(
 		upstreams,
 		(upstream, cb) => {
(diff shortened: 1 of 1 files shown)


src/presets.js
```diff
@@ -150,7 +150,7 @@ const getNestedChangeLog = (
 };
 
 const attachNestedChangelog = (upstreams, commit, callback) => {
-	async.each(
+	async.map(
```
Page-, line 153: Changing this to `eachSeries` would have had the equivalent result but doing everything in serial would have slowed things down a bunch

nazrhom:
@balena-ci rebase

Page-:
@balena-ci rebase

PR #183, opened 2021-02-22
Update balena.yml version

stathismor:
If `balena.yml` exists on a repo, update its version.

Change-type: minor

### DIFF
--- a/src/presets.js
+++ b/src/presets.js
@@ -39,55 +39,55 @@ const octokit = new Octokit({
 
 // prettier-ignore
 const templateDefaults = [
-	'{{#*inline "commits"}}',
-	'{{> render-header }}',
-	'{{> block-newline}}',
-	'{{#each commits}}',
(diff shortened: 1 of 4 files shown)


src/presets.js
```diff
@@ -1125,6 +1125,45 @@ module.exports = {
...
+				return callback(new Error(`Invalid version: ${version}`));
+			}
+
+			const contract = path.join(cwd, 'balena.yml');
+			if (!fs.existsSync(contract)) {
+				return callback(null, version);
+			}
+
+			const content = yaml.safeLoad(fs.readFileSync(contract, 'utf8'));
+			content['version'] = cleanedVersion;
```
nazrhom, line 1162: NIT: this is usually more idiomatic as `content.version = cleanedVersion`, object access with square brackets is usually only used in the context of a variable key, i.e.
```
const key = 'version'
content[key] = cleanedVersion
```

stathismor: Agreed, done.

nazrhom (approved):

PR #186, opened 2021-03-31
Update semserver to do a case-insensitive change-type check

stathismor:
Continuation of #184, updated `semserver` validation and added a couple of tests.

Change-type: minor

### DIFF
--- a/tests/versionist.spec.js
+++ b/tests/versionist.spec.js
@@ -322,11 +322,30 @@ describe('Versionist', function () {
 				.to.throw('No commits to calculate the next increment level from');
 		});
 
-		it('should calculate the next version', function () {
+		it('should calculate the next version lower case', function () {
 			const nextVersion = versionist.calculateNextVersion(
 				[
 					{
 						subject: 'major foo bar',
+					},
+				],
+				{
+					currentVersion: '1.0.0',
+					incrementVersion: _.partial(presets.incrementVersion.semver, {}),
+					getIncrementLevelFromCommit: (commit) => {
+						return _.first(_.split(commit.subject, ' '));
+					},
+				},
+			);
+
+			m.chai.expect(nextVersion).to.equal('2.0.0');
+		});
+
+		it('should calculate the next version titled', function () {
+			const nextVersion = versionist.calculateNextVersion(
+				[
+					{
+						subject: 'Major foo bar',
 					},
 				],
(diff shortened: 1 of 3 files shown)


nazrhom (approved):

tests/versionist.spec.js
```diff
@@ -340,6 +340,25 @@ describe('Versionist', function () {
 
 			m.chai.expect(nextVersion).to.equal('2.0.0');
 		});
+
+		it('should calculate the next version titled', function () {
+			const nextVersion = versionist.calculateNextVersion(
+				[
+					{
+						subject: 'Major foo bar',
```
lurch, line 348: Interesting - does this mean a subject of "Major Tom to Ground Control" would get parsed the same as "Major: Tom to Ground Control"?

stathismor: No @lurch, this is only for the change-type/level.

stathismor: Oh, just noticed your comment again, it's about the colon...  Hm... looks that way...

stathismor: Yeah, you are right, I think the [regex]( https://github.com/product-os/versionist/blob/master/src/presets.js#L451) doesn't care about `:`.

lurch: Hopefully that won't catch people out? :shrug: 
Hmmm, looking at how basic that regex is, I guess that means a subject of "This should only affect a minority of users" would also get interpreted as a `minor` commit??

stathismor: @lurch the regex has now been updated, to check for a case insensitive change-type at the start of the sentence, followed by a colon. So, `Major: Tom to Ground Control` and `major: Tom to Ground Control` will be versioned, but `Tom to major Ground Control` won't.

lurch: Just out of curiosity, does it check (elsewhere) that there's anything after the change-type, or would `major:` by itself be perfectly valid? I don't actually use versionist myself, I'm just trying to help you make sure that any edge-cases / undefined-behaviour are covered :slightly_smiling_face:

stathismor: @lurch No, I don't think there's another check, that I know of. I think someone who just put `major:` as a commit, most probably indented this versioning behavior anyway.

PR #187, opened 2021-04-01
Make subject change-type check stricter

stathismor:
Subject change-type check was very loose before, now the subject needs to start with case-insensitive major|minor|patch, followed by a colon in order for the type to be parsed.

Change-type: patch


nazrhom (approved):

lurch:
Connects to https://github.com/product-os/versionist/pull/186#pullrequestreview-626085919

PR #194, opened 2021-06-21
Add versionist.versionPublishedAt field to package.json

pdcastro:
Change-type: minor

### DIFF
--- a/src/presets.js
+++ b/src/presets.js
@@ -262,6 +262,32 @@ const getCleanFunction = (options) => {
 	return options.clean ? semver.valid : _.identity;
 };
 
+/**
+ * @summary Read the given file and return its contents parsed as JSON
+ * @function
+ * @private
+ *
+ * @param {string} filePath - Path of file to be read and parsed
+ * @param {Function} callback - Error-first callback
+ */
+function readJSON(filePath, callback) {
+	fs.readFile(filePath, 'utf8', (error, data) => {
+		let pjObj;
+		if (error) {
+			if (error.code === 'ENOENT') {
+				error.message = `No such file or directory: ${filePath}`;
+			}
+		} else {
+			try {
+				pjObj = JSON.parse(data);
+			} catch (err) {
+				error = err;
+			}
+		}
+		callback(error, pjObj);
+	});
+}
+
 module.exports = {
 	subjectParser: {
 		/**
@@ -813,18 +839,16 @@ module.exports = {
 
 			async.waterfall(
 				[
-					(done) => {
+					(done) => readJSON(packageJSON, done),
+					(pjObj, done) => {
+						const publishedAt = new Date().toISOString();
 						updateJSON(
 							packageJSON,
 							{
 								version: cleanedVersion,
+								versionist: { ...pjObj.versionist, publishedAt },
 							},
-							(error) => {
-								if (error && error.code === 'ENOENT') {
-									error.message = `No such file or directory: ${packageJSON}`;
-								}
-								return done(error);
-							},
+							done,
 						);
 					},
 					(done) => {
--- a/tests/presets.spec.js
+++ b/tests/presets.spec.js
@@ -25,6 +25,9 @@ const yaml = require('js-yaml');
 const presets = require('../lib/presets');
 
 describe('Presets', function () {
+	const fakeTimeISO = '2021-06-16T12:54:52.000Z';
+	const fakeTimeMs = new Date(fakeTimeISO).getTime();
+
 	describe('.subjectParser', function () {
 		describe('.angular', function () {
 			it('should pass the whole commit as a title when parsing non-angular commits', function () {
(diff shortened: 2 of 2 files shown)


src/presets.js
```diff
@@ -813,18 +831,22 @@ module.exports = {
...
-					(done) => {
+					(done) =>
+						readJSON(packageJSON, (error, pjContents) => {
+							if (error && error.code === 'ENOENT') {
+								error.message = `No such file or directory: ${packageJSON}`;
+							}
+							return done(error, pjContents);
+						}),
+					(pjContents, done) => {
+						const versionPublishedAt = new Date().toISOString();
```
stathismor, line 842: I don't mind this name at all, just wonder if `publishedAt` is enough here.

src/presets.js
```diff
@@ -813,18 +831,22 @@ module.exports = {
...
 						updateJSON(
 							packageJSON,
 							{
 								version: cleanedVersion,
+								balena: { ...pjContents.balena, versionPublishedAt },
```
lurch, line 847: I've got no idea if `versionist` is being used by any non-Balena projects, but if it is should there be a setting to control whether this balena.versionPublishedAt field gets added or not? :shrug:

pdcastro: Interesting consideration - thanks for reviewing it @lurch 👍 

@stathismor, do you know if there is any environment variable that would identify the environment as balena CI?

Alternatively, we could perhaps replace the `balena` namespace with `versionist`:
```
    versionist: { ...pjContents.versionist, versionPublishedAt },
```

By the way @lurch, this PR came from an internal discussion and the motivation is to allow us to determine when (timestamp) a version was published through a query to the npm registry such as:

```
curl https://registry.npmjs.org/balena-cli/12.0.0
```

You'd think that the npm registry would provide a way of concisely extracting that information, but apparently not. The following query does retrieve version timestamps, but for the balena CLI it means downloading ~37MB of JSON data for all 2,500+ historic versions ever published:
```
curl https://registry.npmjs.org/balena-cli
```
The `npm view balena-cli@12.0.0 time` command does the same (~37MB), although it caches the result. I can't rely on caching because I want to prevent ever downloading such a large amount of data to extract a single timestamp.

lurch: I've obviously got no idea of your use-case, but I wonder if you might be able to get the information you're after from the GitHub API instead e.g. https://github.com/product-os/versionist/releases

pdcastro: The use case summary is the balena CLI occasionally checking how long ago a new major version was released, to warn the user about the [deprecation policy](https://github.com/balena-io/balena-cli/blob/master/README.md#deprecation-policy) if it's been more than 6 months (to be confirmed) that a new major version was released.

Indeed the GitHub API provides this information, but we thought it was not the best solution because the request would not be authenticated and would be subject to GitHub's rate limit of [60 requests per hour per origin IP address](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting). We were worried that requests would be turned down too often at some sites.

lurch: Oh, so this is _only_ useful for balena-cli, and not for all javascript packages updated by `versionist`? Feels a bit overkill to add a feature to versionist, that'll only be used by a single package? :shrug: Seems like all you need is a `balena-cli-release-dates.json` hosted on some static webserver somewhere?
But of course it's your ecosystem, so you're free to do whatever you want! :slightly_smiling_face:

pdcastro: Not necessarily, a "date of publishing" field like `versionist.publishedAt` (an improvement over `balena.versionPublishedAt`, thanks to the review comments!) is a generic feature that could be used by any package for any purpose with a query to the npm registry at `https://registry.npmjs.org/<package-name>/<version>`. The balena CLI use case is our present motivation for the PR.

lurch: If it's a generally-useful-feature, I wonder if it might be worth adding it to the rust and python presets too? :shrug:

pdcastro: Possibly, but this PR was motivated by a feature and limitation of the npm registry: that the registry allows fetching the `package.json` file for a published version (`https://registry.npmjs.org/balena-cli/12.0.0`), while not allowing a fine grained query to fetch the date when that version was published (without having to download the whole version history that could be tens of megabytes).
I don't know enough about the Rust and Python "registries" to know that they have a similar limitation, and I don't know enough about the Rust and Python packaging files (`cargo.toml`, `__init__.py`) to be confident that adding a `versionist` section to those files won't cause problematic side effects.
Adding the feature only to the npm preset in this PR does not prevent the feature being added to the other presets later on, when an actual use case was identified.

src/presets.js
pdcastro: Done. `versionist.publishedAt` is also a bit clearer than `balena.publishedAt` in relation to the fact that the "publishing" refers to a new version of the package.

pdcastro:
The last force push was to amend the commit message, which now reads:
```
Add versionist.publishedAt field to package.json when updating the version

Change-type: minor
```

nazrhom (approved):

PR #198, opened 2021-10-21
Use balena-semver instead of resin-semver

bartversluijs:
The package `resin-semver` is being used, but it's deprecated.
Replaced it with the newer `balena-semver`.


klutchell:
Thanks for the PR, I've merged the change in https://github.com/product-os/versionist/pull/200 to avoid the conflicts!

Cheers!

PR #199, opened 2021-11-04
Update nested package-lock version

stathismor:
Updates the nested `packages[""].version` in the same way as the root version. I don't know if this key will exist in following `lockfileVersion`s, there is an explicit check for it. Also not sure under what node version this code will be run (need to update a couple of other packages that have this dependency) so check is not using the optional chaining operator (?.).

Change-type: minor

### DIFF
--- a/src/presets.js
+++ b/src/presets.js
@@ -865,6 +865,36 @@ module.exports = {
 							},
 						);
 					},
+					(done) =>
+						readJSON(packageLockJSON, (readError, pljObj) => {
+							if (readError && readError.code === 'ENOENT') {
+								return done(null);
+							}
+
+							if (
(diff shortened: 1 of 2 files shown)


src/presets.js
```diff
@@ -39,55 +39,55 @@ const octokit = new Octokit({
 
 // prettier-ignore
```
Page-, line 40: It looks like you've reformatted this in a way that ignores this `prettier-ignore` - the reason for the custom formatting is that the strings are in essence code which is nicely indented for readability and the auto-formatting has clobbered that readability now

stathismor: Sure, I can change that. I just ran `npm run prettify` which does not seem to respect the ignore...

Page- (approved):

PR #202, opened 2021-11-23
preserve YAML document when writing version

Hades32:
before, comments were just removed.

This required a change from `js-yaml` to `yaml`, but AFAICT the latter is even more complete, so that shouldn't be an issue


stathismor (commented):
LGTM!

Hades32:
@balena-ci I self-certify!

PR #203, opened 2021-11-24
drop update-json dependency

Hades32:
after 3 years 1.0.1 was not published and it's just a single file


Hades32:
@balena-ci I self-certify!

PR #205, opened 2021-12-17
Get changelog.yml blob

stathismor:
GitHub get content API does not allow pulling files bigger than 1MB, switching to using trees and blob. 

Change-type: patch

### DIFF
--- a/src/github.ts
+++ b/src/github.ts
@@ -1,0 +1,59 @@
+import { Octokit } from '@octokit/rest';
+
+/**
+ * @summary Get CHANGELOG.yml
+ * @function
+ * @public
+ *
+ * @param {string} owner - repo owner
+ * @param {string} repo - repo
+ * @param {octokit} octokit - Octokit client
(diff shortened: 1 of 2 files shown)


src/github.ts
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * Get the CHANGELOG.yml file using the trees and blob APIs. getContent API
+ * does not allow for files bigger than 1MB so we need to get the sha and fetch it.
+ */
+export const getChangelogYML = async (
+	owner: string,
+	repo: string,
+	branch: string,
+	octokit: Octokit,
+): Promise<any> => {
```
Page-, line 22: I assume `blobResponse.data.content` is a buffer?
```suggestion
): Promise<Buffer> => {
```
But if not then presumably it's a string:
```suggestion
): Promise<string> => {
```

stathismor: It's a string @Page- , done!

stathismor:
@balena-ci I self-certify!

PR #208, opened 2022-01-17
Add changelog title support

stathismor:
versionist can accept a title (--title, -t). It won't do anything with it by default, but it is accessible in the template.

Change-type: minor


stathismor:
@balena-ci I self-certify!

PR #210, opened 2022-03-02
Deduce submodule update versions automatically

alexgg:
The nested changelog support currently needs to be provided with the
version range on the commit body, similarly to:

  Update <submodule> from <initialVersion> to <targetVersion>

This is cumbersome and does not work well with tools that automatically
update submodule dependencies like renovatebot.

This commit keeps the legacy behaviour but will extract the version range
from git metadata when not provided. This new behaviour can be used by
specifyng a commit body similar to:

  Update <submodule>

Change-type: patch
Signed-off-by: Alex Gonzalez <alexg@balena.io>

### DIFF
--- a/src/presets.js
+++ b/src/presets.js
@@ -32,6 +32,7 @@ const markdown = require('./markdown');
 const yaml = require('yaml');
 const execSync = require('child_process').execSync;
 const { Octokit } = require('@octokit/rest');
+const childProcess = require('child_process');
 
 const octokit = new Octokit({
 	debug: Boolean(process.env.DEBUG),
@@ -150,6 +151,32 @@ const getNestedChangeLog = (
 		});
 };
 
+const resolveTag = (shash, upstream) => {
+	const result = childProcess.spawnSync(
+		'git',
+		['describe', '--abbrev=0', shash],
+		{
+			cwd: `${process.cwd()}/${upstream}`,
+		},
+	);
+
+	return result.stdout.toString().trim();
+};
+
+const processSubmoduleOutput = (data, upstream) => {
+	var lines = `${data}`.split('\n');
+	for (let i = 0; i < lines.length; i++) {
+		const line = lines[i];
+		const regexp = new RegExp(`Submodule ${upstream} (\\S+)\\.\\.(\\S+):`);
+		const match = line.match(regexp);
+		if (match) {
+			const previousVersion = resolveTag(match[1], upstream);
+			const newVersion = resolveTag(match[2], upstream);
+			return [previousVersion, newVersion];
+		}
+	}
+};
+
 const attachNestedChangelog = (upstreams, commit, callback) => {
 	async.map(
 		upstreams,
@@ -168,6 +195,32 @@ const attachNestedChangelog = (upstreams, commit, callback) => {
 					targetVersion,
 					cb,
 				);
+			}
+			const regexp1 = new RegExp(`[Uu]pdate ${upstream.pattern}`);
+			const match1 = commit.body.match(regexp1);
+			if (match1) {
+				const result = childProcess.spawnSync('git', [
+					'log',
+					'--oneline',
+					'--submodule',
+					'-U0',
+					'HEAD^1..HEAD',
+				]);
+				const response = processSubmoduleOutput(
+					result.stdout,
+					`${upstream.pattern}`,
+				);
+				if (response) {
+					const currentVersion = response[0];
+					const targetVersion = response[1];
+					return getNestedChangeLog(
+						upstream,
+						commit,
+						currentVersion,
+						targetVersion,
+						cb,
+					);
+				}
 			}
 			return cb(null);
 		},


alexgg:
Testing details:
* Fork a device repository on a test github account
* Install renovate on it
* Onboard and configure renovate with https://github.com/balena-os/renovate-config
* Fork versionist and deploy and [npm package](https://www.npmjs.com/package/ag-versionist) with my modifications
* Fork `balena-versionist` modify it to depend on `ag-versionist` and deploy an [npm package](https://www.npmjs.com/package/ag-balena-versionist)
* Fork versionist github action and modify it to use the npm packages releases above
* Install the action in the device repository
* Configure the action with:
```
name: Run versionist
on:
  push:
    branches:
      - master

jobs:

  versionist:
    name: Run versionist
    if: "!contains(github.event.head_commit.author.name, 'versionist')"   # Ignore push events made by the service account
    runs-on: ubuntu-latest
    outputs:                                              # (optional) Only if you want to use them in next jobs
      version: ${{ steps.versionist.outputs.version }}    # version: project's version after running versionist
      updated: ${{ steps.versionist.outputs.updated }}    # updated: true if the version has been updated
    steps: 
    - name: Checkout project
      uses: actions/checkout@v2
      with:
        submodules: recursive
        fetch-depth: 0                                    # We need all commits and tags
        persist-credentials: false                        # Next step needs to use service account's token
    - name: Run versionist
      id: versionist                                      # (optional) Only needed if using outputs
      uses: alexgg/versionist@alexgg-nested-changelogs
      with:
        # Provide your versionist service account details
        github_email: 'yoctocookbook2ndedition@lindusembedded.com'
        github_username: 'yoctocookbook2ndedition'
        github_token: ${{ secrets.GH_VERSIONIST_TOKEN }}
```

With the above, the device repostiory dependencies are updated and the changelog is nested correctly.

src/presets.js
```diff
@@ -169,6 +198,30 @@ const attachNestedChangelog = (upstreams, commit, callback) => {
 					cb,
 				);
 			}
+			const regexp1 = new RegExp(`[Uu]pdate (layers)?/${upstream.pattern}`);
+			const match1 = commit.body.match(regexp1);
+			if (match1) {
+				const result = childProcess.spawnSync('git', [
```
stathismor, line 204: Maybe from here and further down, it could be in its own function.

stathismor (commented):
Looks good, but I have one concern. `layers` is just a `balena` thing, right? `versionist` is supposed to be generic and run on every repo that uses it right. Even thought it does not break existing behaviour, I wonder if we're putting a very specific feature on a generic tool. And whether all that code should lie in `versionist.conf.js`.

src/presets.js
alexgg: thanks @stathismor you're right that `layers` is balena specific.

Not sure how to approach this from `versionist.conf.js`, could you maybe explain your thoughts?

Another way to approach it would be to modify the way the upstream repositories are defined in `repo.yml` as currently we only state the submodule name and not the path, as in:
```
upstream:                                                                                               
- repo: 'meta-balena'                                                         
    url: 'http://github.com/balena-os/meta-balena'                              
- repo: 'balena-yocto-scripts'                                                
    url: 'http://github.com/balena-os/balena-yocto-scripts'
```
Instead of:
```
- repo: 'layers/meta-balena'                                                         
    url: 'http://github.com/balena-os/meta-balena'    
```
This would involve changing all the current `repo.yml`, and I am not sure if that change would affect anything else.

Finally, another solution not that intrusive is using a broader regex and still relying on the submodule name for identification:
```
const regexp1 = new RegExp(`[Uu]pdate (.*)?${upstream.pattern}`);
```
What do you think?

stathismor: 1.
> Not sure how to approach this from versionist.conf.js, could you maybe explain your thoughts?

Yeah, sure. You know that repos can have [custom versionist configs](https://github.com/product-os/versionist#configuration), right? An example how that can look [here](https://github.com/product-os/balena-versionist/blob/master/lib/repo-type-mappings/yocto-layer/versionist.conf.js). A `versionist.conf.js` at the repo's root, would overwrite any default `versionist`/`balena-versionist` behaviour. Looking at it now, the relevant option here I think is `transformTemplateDataAsync`. Current code does:
```
	transformTemplateDataAsync: {
		'nested-changelogs': (options, data, callback) => {
			async.map(
				data.commits,
				_.partial(attachNestedChangelog, options.upstream),
				(err, commits) => {
					if (err) {
						return callback(err);
					}
					data.commits = _.sortBy(commits, (commit) => {
						return Boolean(commit.nested);
					});

					return callback(err, data);
				},
			);
		},
	},
```
You could copy that on the conf file, along with your own implementation of `attachNestedChangelog`, as you have here.

If it's just one or two repos that want this behavior, the conf file would be on their root. If we want to have this for all repos of a type, it could go to `balena-versionist` configurations https://github.com/product-os/balena-versionist/tree/master/lib/repo-type-mappings

2. Regarding changing upstream names or having a more generic regexp. We're still using `repo.yml` though which is a balena thing... Also, wouldn't having a generic regexp increase the chances of confusing the tool, when someone just writes `Update <submodule name>`?

alexgg: thanks @stathismor I understand now how to approach this from `balena-versionist` and `versionist.conf.js`.

However, the reason why I decided to put this into versionist initially is because it is a usage improvement. Right now you have to manually craft a commit body to link the nested changelog behaviour, and you have to manually type the current and new versions, and you can get it wrong and there is no way back. But versionist has all the information to extract those versions for you, you only have to match the submodule name in the commit log with the one defined in `repo.yml`.

Not sure that `repo.yml` is a balena thing. I don't think you can use nested changelogs if you don't define the upstreams in `repo.yml`, so we either agree that nested changelogs is balena-specific, or `repo.yml` is not.

In summary,

* I can refactor the change into `versionist.conf.js, making it balena specific
* We can either change the upstreams in `repo.yml` or extend the regex.

About extending the regex, not sure what you mean by:

> confusing the tool, when someone just writes Update <submodule name>

The tool will try to extract nested changelogs in that case, but if there are no upstreams defined in `repo.yml` or not submodule data, it will not find nested data and return empty.

Anyway, I really just want to solve this for device repositories, so whatever you guys decide is the path forward is fine with me.

alexgg: I think I understand why a more generic regex is messy, so it's between moving it to `versionist.conf.js` or modifying the upstreams in `repo.yml`.

Thinking about it, I think that it's a usage improvement on the nested changelog functionality already in `versionist` so I'd like to understand what are the drawbacks in changing the upstreams in `repo.yml`

So, I am tending to try the following:
Leave the regex match as:
```
const regexp1 = new RegExp(`[Uu]pdate ${upstream.pattern}`);
```
And change the `repo.yml` in device repositories to match as in:
```
- repo: 'layers/meta-balena'                                                         
    url: 'http://github.com/balena-os/meta-balena'
- repo: 'balena-yocto-scripts'                                                
    url: 'http://github.com/balena-os/balena-yocto-scripts'
```
What do you think?

stathismor: > Not sure that repo.yml is a balena thing. I don't think you can use nested changelogs if you don't define the upstreams in repo.yml, so we either agree that nested changelogs is balena-specific, or repo.yml is not

Fair point, I missed that nested changelogs do use `repo.yml` anyway.

Ok, I am fine with changing upstreams :+1:

alexgg: I have now removed `layers` and just use the defined `upstream` from `repo.yml`. I have updated my test setup and been testing with a few submodule PRs, all looks good.
I expect that most projects are using submodules at the project root and not in a submodule, so only the device repositories will need to update `repo.yml`.

src/presets.js
```diff
@@ -169,6 +196,32 @@ const attachNestedChangelog = (upstreams, commit, callback) => {
 					cb,
 				);
 			}
+			const regexp1 = new RegExp(`[Uu]pdate ${upstream.pattern}`);
```
stathismor, line 199: Let's rename `regexp1` and `match1` to `regexp` and `match`.

src/presets.js
```diff
@@ -150,6 +151,32 @@ const getNestedChangeLog = (
...
+			cwd: `${process.cwd()}/${upstream}`,
+		},
+	);
+
+	return result.stdout.toString().trim();
+};
+
+const processSubmoduleOutput = (data, upstream) => {
+	var lines = `${data}`.split('\n');
+	for (var i = 0; i < lines.length; i++) {
```
stathismor, line 168:
```suggestion
	for (let i = 0; i < lines.length; i++) {
```

src/presets.js
```diff
@@ -150,6 +151,32 @@ const getNestedChangeLog = (
...
+		let line = lines[i];
```
stathismor, line 169:
```suggestion
		const line = lines[i];
```

stathismor (commented):
Minor comments, LGTM

src/presets.js
alexgg: @stathismor we already have `regexp` and `match` const above in this function so they cannot be redefined unless we don't use consts. Any other variable name suggestion you prefer?

src/presets.js
alexgg: Yep, safer as const.

alexgg:
@balena-ci I self-certify!

PR #232, opened 2022-10-19
Fix generate-changelog.sh script when there are no versioned PRs

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #243, opened 2023-05-02
Update nested npm-shrinkwrap version

pipex:
Updates the nested packages[""].version in the same way as the root version. We already do this for package-lock.json

Change-type: minor
Relates-to: #199


pipex:
TODO: change [src/presets.js](src/presets.js) to allow tests to pass

pipex:
I self-certify!

PR #249, opened 2023-10-30
Fail changelog generation if not authorized

pipex:
By default the `curl` command will not return an error reponse for non 2xx HTTP response. This change adds the [-f flag](https://curl.se/docs/manpage.html#-f) to the pull request query to allow it to fail if authorization fails instead of looping forever.

Change-type: patch


klutchell:
LGTM

PR #257, opened 2025-03-04
Balena semver 3

thgreasi:
See: https://balena.zulipchat.com/#narrow/channel/346007-balena-io.2FbalenaCloud/topic/balenaMachine.20repo.20stuck.20on.20v14.2E1.2E1000/near/503081215
Change-type: major

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,9 @@
   "bin": {
     "versionist": "./bin/cli.js"
   },
+  "engines": {
+    "node": ">=20"
+  },
   "scripts": {
     "pretest": "npm run prepack && npx tsc --noEmit --project tsconfig.dev.json",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -12,6 +12,9 @@
   "bin": {
     "versionist": "./bin/cli.js"
   },
+  "engines": {
+    "node": ">=20"
```
thgreasi, line 16: balena-semver requires node 16, but that's already EOL so I decided to go to 20.
Let me know if we need to lower in case it makes it harder to bump versionist in other dependent packages.

klutchell: Technically it's not even a dependency of [balena-versionist](https://github.com/product-os/balena-versionist/blob/master/package.json) right now, so it should be fine?

klutchell (approved):

